PR #1, opened 2019-06-19
Moving core20 snap to $devel/$nextlts series

xnox:
Build on eoan distro-series locally with $ sudo snapcraft. And fix things.

- Makefile
  - Try to use $devel daily tarball, until we hit $nextlts.
  - Use $nextlts once we do.
- snapcraft.yaml:
  - make version a string
  - declare bogus core18 base (core20 or base don't work)
  - add distro-info build-dep
- rename most includes from `/lib` to `/usr/lib` as we now have merged-usr
- add kvm group 112 *not sure if this was not used before for anything else*
- clean up private dir, newly created by tmpfiles

At the end of the build snapcraft blows up. So don't merge this yet.


snapcraft.yaml
```diff
@@ -1,10 +1,15 @@
...
+version: "20"
 summary: Runtime environment based on Ubuntu 20.04
 description: |
   The base snap based on the Ubuntu 20.04 release.
 confinement: strict
 type: base
+## Self-reference much?! =)
+## The plugin used by part 'boostrap' does not support snaps using base 'core20'.
+#base: core20
+base: core18
```
mvo5: Build-base maybe if thats supported yet by snapcraft?

hooks/000-provide-uids-gids.chroot
```diff
@@ -267,6 +268,7 @@ systemd-timesync:!::
 systemd-network:!::
 systemd-resolve:!::
 systemd-bus-proxy:!::
+kvm:!::
```
mvo5: Why do we need the kvm user?

xnox: See the commit message....

```
Add kvm system group  …
udev postinst adds system group kvm to ensure that /dev/kvm is owned
by the right group. There is no pre-allocated fixed groupid for
it.

Adding a new allocation kvm 112.

However, if that allocation was used in core16 before a different
(higher) one should probably be used.
```

xnox: and it's not user, but group.

xnox:
@cmatsuoka @mvo5 this is now ready to be merged. Lots of test snaps got build in https://launchpad.net/~xnox/+snap/core20 and tested by cmatsuoka and everything seems to work.

This will switch core20 snap to use eoan.

mvo5:
@xnox Looks like travis is unhappy right now, I wonder if we can fix this somehow?

cmatsuoka:
Add `--use-lxd` (or `--destructive-mode`) to snapcraft when building the snap. Possibly `TESTDIR` in the Makefile will have to be adjusted as well.

xnox:
@mvo5 @cmatsuoka it's now green!

PR #10, opened 2019-10-12
Provide lxd group by default

### LINKED ISSUE #9: core20 groups are odd and incompatible with core16
1. group 113 is docker
1. group 113 should not be kvm
1. kvm group should be something else
1. lxd group needs to be added

xnox:
Fixes #9


xnox:
@mvo5 I still want to understand the large set of groups that are baked into the core snaps, and how addition / renumbering of those affect re modelling.

PR #13, opened 2019-11-19
handle-writable-paths: extract the writable-path handling

mvo5:
This PR moves the code that handle the writable-path for a core
system from the initramfs to the core20 snap. This has the benefit
that we don't have to change the initramfs if this code needs to
change. The initramfs will just call it from the core20 snap.


xnox:
Not sure why we care about writing a valid fstab.
But this should do for now.
And ideally I would want to move all the mount calls to systemd-mount, such that there are units created for all of them.

PR #14, opened 2019-11-19
run-snapd-from-snap: check for snapd.service existing too

mvo5:
If state.json exists, but snapd.service doesn't, then we were likely
interrupted before the snapd-seeding from /tmp was able to bootstrap the
full snapd from the snap, so we still need to do that now.

Fixes: https://bugs.launchpad.net/snapd/+bug/1845310

Signed-off-by: Ian Johnson <ian.johnson@canonical.com>


mvo5:
Any chance this could get reviewed/merged?

PR #15, opened 2019-11-27
Provide ubuntu-seed mount

xnox:


xnox:
I'm not sure, but I think this is what we need to do upon initialization.... at least for the recover mode.

mvo5:
Yes, this works,let's do it!

mvo5:
Will you do this differently or why is it closed? Or should I reopen the code that does it in snap-bootstrap initramfs-mounts ?

mvo5:
Should https://github.com/snapcore/snapd/pull/7797 be reopened?

xnox:
Merged as a rebase....

PR #16, opened 2019-12-13
bootstrap from locally build snapd snaps too

mvo5:
An unasserted snapd snap will be located in
```
/var/lib/snapd/seed/systems/$label/snaps
```
which the current run-snapd-from-snap bootstrap code will not
find. This commit fixes this to make this work short-term.

However we will need to write the right snapd snap into the
grubenv/modeenv so that when there are multiple recovery systems
there is no confusion.

### DIFF
--- a/static/usr/lib/core20/run-snapd-from-snap
+++ b/static/usr/lib/core20/run-snapd-from-snap
@@ -7,7 +7,8 @@ set -eux
 
 # run_on_unseeded will mount/run snapd on an unseeded system
 run_on_unseeded() {
-    SEED_SNAPD="$(find /var/lib/snapd/seed/snaps/ -name "snapd_*.snap")"
+    # XXX: read the snapd snap to bootstrap from via /var/lib/snapd/modeenv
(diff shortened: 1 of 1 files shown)


static/usr/lib/core20/run-snapd-from-snap
```diff
@@ -7,7 +7,8 @@ set -eux
 
 # run_on_unseeded will mount/run snapd on an unseeded system
 run_on_unseeded() {
-    SEED_SNAPD="$(find /var/lib/snapd/seed/snaps/ -name "snapd_*.snap")"
+    # XXX: read the snapd snap to bootstrap from via /var/lib/snapd/modeenv
+    SEED_SNAPD="$(find /var/lib/snapd/seed/ -name "snapd_*.snap")"
```
pedronis: couldn't/would't snapd be already mounted via snap-bootstrap ?

xnox: @pedronis at the moment snap-bootstrap does not mount snapd. Imho, it should be validating and mounting snapd. As otherwise, one can freeze execution, and inject unverified snapd before this line of code runs.

PR #18, opened 2020-01-06
static: try using /run/mnt/snapd first in run-snapd-from-snap

mvo5:
With https://github.com/snapcore/snapd/pull/7957 snap-boostrap
will mount the "correct" snapd snap to /run/mnt/snapd. With that
in place we can simplify the run-snapd-from-snap script and also
avoid potentially mounting a random snapd snap when there are
multiple recovery systems.

This commit keeps the old way of doing things for compatiblity
but once the snapd PR is merged we can remove the compat code.


anonymouse64:
LGTM

xnox:
Also will need patches to the initrd, to keep the /run/mnt/snapd around, without unit dependencies it will be tore down by pivot-root =(

PR #19, opened 2020-01-08
static: add /etc/environment to writable-paths

mvo5:
Ogra reported that /etc/environment is not writable on uc20 based
systems. This commit fixes this. We also noticed this in our spread test.


xnox:
Was merged in core18 already 2 months ago.

xnox:
core18 commit https://github.com/snapcore/core18/commit/8fb8aff67a0f11b9c1b27cdc796902e8b58488aa#diff-472f335517681ecd7b942a5096a423bb

PR #23, opened 2020-02-24
hooks: ensure console-conf@ is started after snapd.recovery-chooser-trigger.service

bboozzoo:
Combined with `Type=oneshot` of the chooser trigger service, this ensures that by
the time console-conf runs, the chooser exited and optionally left a marker
file.

NOTE: this makes the boot process longer by the amount of time it takes the chooser trigger to timeout.

Related snapd PR: https://github.com/snapcore/snapd/pull/8156

### DIFF
--- a/hooks/201-console-conf-after-chooser.chroot
+++ b/hooks/201-console-conf-after-chooser.chroot
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -e
+
+# Ensure that console-conf is started only once the chooser detection has completed
+echo "Fix console-conf systemd unit to run after snapd recovery chooser trigger detection"
+for f in console-conf@.service serial-console-conf@.service; do
+    p="/lib/systemd/system/$f"
+    sed -i 's/After=\(.* core.start-snapd.service\)/After=\1 snapd.recovery-chooser-trigger.service/' "$p"
+    # ensure sed worked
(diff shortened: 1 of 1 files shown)


anonymouse64:
I understand that it might be easier to fix this here, but couldn't we just fix this upstream at https://github.com/CanonicalLtd/subiquity/blob/master/debian/console-conf.console-conf%40.service?

(also I furthermore understand that this hack has been inherited by https://github.com/snapcore/core18/pull/74 and https://github.com/CanonicalLtd/subiquity/pull/389 which only seems to have targeted bionic and never made it's way up to focal, so perhaps the thing to do is just to keep hacking here, but it feels unfortunate we can't get this upstream instead)

xnox:
Yeap, please resubmit against:
https://github.com/CanonicalLtd/subiquity/blob/master/debian/console-conf.console-conf%40.service
https://github.com/CanonicalLtd/subiquity/blob/master/debian/console-conf.serial-console-conf%40.service

I know the levels of indirection here are not even funny.

bboozzoo:
> I know the levels of indirection here are not even funny.

:laughing: 

Will open a PR to subiquity master. I'll include the core18 bit too.

bboozzoo:
This is done now in https://github.com/CanonicalLtd/subiquity/pull/638

PR #24, opened 2020-02-24
hooks: delete console-conf hack

anonymouse64:
This was always meant to be temporary while it was upstreamed, however that didn't happen, but now is happening.

Note: this should not be merged until https://github.com/CanonicalLtd/subiquity/pull/637 is merged and released, otherwise console-conf will break on UC20.


anonymouse64:
Closed in favor of https://github.com/snapcore/core20/pull/25

Issue #27, opened 2020-03-20
iptables is missing


alfonsosanchezbeato:
iptables and related libraries are missing from core20. Note that the `firewall-control` interface grants being able to call the iptables binary from core.

xnox:
does it need to be iptables? nftables? either?

alfonsosanchezbeato:
I'd say iptables, that is what the snapd interface gives permissions for. Also, iptables seems to be still the default in focal, at least I have it installed, but no nftables.

PR #28, opened 2020-03-23
Add iptables package

alfonsosanchezbeato:
iptables is used by snapd's firewall-control interface. Fixes
https://github.com/snapcore/core20/issues/27.


alfonsosanchezbeato:
I only see iptables by default in focal, not nftables. If we want the later, we would need to modify the snapd interfaces too.

zyga:
I understand that we'd have to change interfaces and perhaps also the confinement system. My only worry was for the consistency of Ubuntu 20.04 with Ubuntu Core 20.

PR #33, opened 2020-04-03
writable-path: make /etc/machine-id writable

mvo5:
This should ensure we have a correct persistent journal.


bboozzoo:
Verified this with core20 snap buit from this PR. Looks like it works fine, /etc/machine-id is generated and persists across reboots.

PR #39, opened 2020-04-14
ensure that /host exists

bboozzoo:
The mounts from the host system will be placed under the /host directory.

### DIFF
--- a/hook-tests/020-extra-files.test
+++ b/hook-tests/020-extra-files.test
@@ -2,3 +2,5 @@
 
 test -d /snap
 test -d /var/snap
+
+test -d /host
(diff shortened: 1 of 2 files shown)


hook-tests/020-extra-files.test
```diff
@@ -2,3 +2,5 @@
 
 test -d /snap
 test -d /var/snap
+
+test -d /host
```
jhenstridge, line 6: This should probably be `test -d host`: you want to test the contents of the prime directory rather than what's on the build host.  The other tests in the look like they pass by chance since they happen to exist on the build host.

bboozzoo: Duh, I think you're right. I'll push a fix for the whole test.

bboozzoo: @jhenstridge thanks for spotting that! I've opened a PR with the fix here: https://github.com/snapcore/core20/pull/40

PR #45, opened 2020-04-17
Add arches

xnox:


mvo5:
Thanks for this PR. It looks great, but unfortunately CI seems to be not quite working.

anonymouse64:
This may work better if it is rebased on master to get https://github.com/snapcore/core20/pull/74 incorporated here

xnox:
> This may work better if it is rebased on master to get #74 incorporated here

I can try, but previously any "snaps:" stanzas would fail to install any snaps. Installing them twice in a row would make them work. But overall it seemed impossible to use snaps on arm64 "just like on amd64". But hopefully enough time has passed to have all of that fixed.

xnox:
doesn't work, whatever.

PR #46, opened 2020-04-21
static: add new handle_writable_defaults() to handle-writable-paths

mvo5:
The new handle_writable_defaults() helper will take the defaults
from $writable/system-data/_writable_defaults/ and apply them on top of
the populated $writable/system-data.

This will allow providing image defaults in a clean way.

This is https://github.com/snapcore/core-build/pull/59 for UC20.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,9 @@ test:
 				exit 1; \
 			fi; \
 	    	done; \
+	set -ex; for f in $$(pwd)/tests/test_*.sh; do \
+		sh -e $$f;
+	done;
 
 # Display a report of files that are (still) present in /etc
 .PHONY: etc-report
--- a/Makefile
+++ b/Makefile
@@ -75,8 +75,8 @@ test:
 			fi; \
 	    	done; \
 	set -ex; for f in $$(pwd)/tests/test_*.sh; do \
-		sh -e $$f;
-	done;
+		sh -e $$f; \
+	done
 
 # Display a report of files that are (still) present in /etc
 .PHONY: etc-report
--- a/static/usr/lib/core/handle-writable-paths
+++ b/static/usr/lib/core/handle-writable-paths
@@ -158,7 +158,7 @@ handle_writable_defaults()
         dstpath="${rootmnt}/writable/system-data"
         for fileordir in "$srcpath"/* ; do
             # skip empty $srcpath/
-            [ ! -e "$fileordir" ] && continue
+            [ ! -e "$fileordir" ] && [ ! -L "$fileordir" ] &&  continue
 
             # copy
             cp -a "$fileordir" "$dstpath"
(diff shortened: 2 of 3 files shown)


mvo5:
This is not quite right yet

static/usr/lib/core/handle-writable-paths
```diff
@@ -141,6 +141,29 @@ handle_writable_paths()
...
+{
+        srcpath="${rootmnt}/writable/system-data/_writable_defaults"
+        if [ ! -d "$srcpath" ] || [ -e "${srcpath}/.done" ]; then
+                return
+        fi
+
+        dstpath="${rootmnt}/writable/system-data"
+        for fileordir in "$srcpath"/* ; do
+            # skip empty $srcpath/
+            [ ! -e "$fileordir" ] && [ ! -L "$fileordir" ] &&  continue
```
sil2100, line 161: I guess I don't understand this line of code. We're handling empty srcpath, which -e makes sense, but then if the path does not exist we also check if it is not a symlink? Will the second conditional ever be true when there's an && between?

static/usr/lib/core/handle-writable-paths
```diff
@@ -141,6 +141,29 @@ handle_writable_paths()
 			continue
 		fi
 	done
+
+        # now apply the defaults on top
+        handle_writable_defaults
```
sil2100, line 146: I think these two lines have inconsistent indenting (using spaces when rest uses tabs).

static/usr/lib/core/handle-writable-paths
mvo5: This is to catch the case when there is a broken symlink. In this case we still want to copy the broken symlink (as it may point to something sensible later when the full core20 writable-path stuff is assembled). There is a testcase for this in the new test (that checks if broken symlinks work).

sil2100: Ah, ok, makes sense! Forgot that -e follows symlinks. Thanks for reminding me of that!

Makefile
```diff
@@ -74,6 +74,9 @@ test:
 				exit 1; \
 			fi; \
 	    	done; \
+	set -ex; for f in $$(pwd)/tests/test_*.sh; do \
+		sh -e $$f; \
+	done
```
xnox, line 79: Why `-e` twice and `-x` only once?

mvo5: The global "-ex" ensures we see what test is run. The sh -e just ensures the individual test fails when it should fail. Note that the test ifile tself adds a "-x/+x" in the test section to make sure that failures are visible. It's a very poor mans testsuite and I would be happy to replace it one day but for now I think it's sufficient.

static/usr/lib/core/handle-writable-paths
mvo5: thanks, fixed and pushed

PR #47, opened 2020-04-21
hooks: use explicit /sbin/init to run systemd multi-user.target

anonymouse64:
For some reason, systemd will no longer automatically run to the multi-user.target anymore, and instead stops doing things right at the basic.target, which has broken uc20 booting.

There is probably a more elegant way to do this with symlinks and stuff, but this is clean and understandable.


xnox:
So initrd did something weird. if this is now needed. Checking things.

PR #49, opened 2020-04-22
fix broken symlinks in /etc/writable

mvo5:
There is a bugreport in the forum that there are some broken
symlinks in /etc that should point to /etc/writable. This PR
create placeholder files and makes the symlinks relative.

This is https://github.com/snapcore/core18/pull/90 for UC20.


tests/final/etc-writable-symlinks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
+
+echo "Test that the symlinks for etc/{timezone,localtime,hostname} exist"
+set -x
+for f in timezone localtime hostname; do
```
xnox, line 7: i think at least timezone should be a symlink to the UTC timezone that should be available in the core20 snap.

mvo5: Thanks, I updated the 030-fix-timedatectl.chroot which already sets /etc/timezone (but not localtime). This should be ready for a re-review now.

mvo5:
Travis breaks right now because tests on master are broken. Once #65 is merged this should be fine.

xnox:
Having sensible defaults for timezone stuff is good. Having empty target for writable too. I am concerned about https://github.com/snapcore/core20/issues/61 but we can solve that separately.

I wonder if we should be setting hostname to something random, or like just "uc20" or "ubuntu". But that can be done separately. I.e. it can come over dhcp.

PR #54, opened 2020-05-07
static: switch /etc/cloud from synced to persistent

mvo5:
This commit move /etc/cloud from a synced dir to a persistent
dir. Synced dirs have strange semantics, e.g. the user cannot
remove a file that is also present in the core20 snap in the
dir without it coming back on the next reboot. Hence we want
to avoid any synced dirs.

This requires the following prereqs:
https://github.com/snapcore/core20/pull/46
https://github.com/snapcore/snapd/pull/8612

first.

### DIFF
--- a/static/etc/system-image/writable-paths
+++ b/static/etc/system-image/writable-paths
@@ -52,12 +52,7 @@
 /var/lib/dbus                           auto                    persistent  none        none
 /var/lib/dhcp                           auto                    persistent  none        none
 # cloud-init
-# FIXME: we currently write an empty /writable/system-data/etc/cloud directory
-#        when creating the image.  So setting /etc/cloud to  "persistent"
-#        instead of "synced" will result all writes getting skipped. Once
-#        the image does no longer contains this empty dir this patch can be
-#        reverted.
-/etc/cloud                              auto                    synced      none        none
+/etc/cloud                              auto                    persistent  none        none
 /var/lib/cloud                          auto                    persistent  none        none
 # for various clouds like GCE
 /etc/sysctl.d                           auto                    persistent  transition  none


xnox:
I see that https://github.com/snapcore/snapd/pull/8612 is merged, but does it also need to be made available in the beta channel too, before we merge this in core20?

mvo5:
Yeah, I think we want snapd with the right code to hit at least beta (probably stable) first to avoid any out-of-syncness here. It's not urgent to merge this we just need it before 1.0

anonymouse64:
FWIW, the required bits from snapd will be landing in 2.45.3 for this, see https://github.com/snapcore/snapd/pull/9052

static/etc/system-image/writable-paths
```diff
@@ -52,12 +52,7 @@
 /var/lib/dbus                           auto                    persistent  none        none
 /var/lib/dhcp                           auto                    persistent  none        none
 # cloud-init
-# FIXME: we currently write an empty /writable/system-data/etc/cloud directory
-#        when creating the image.  So setting /etc/cloud to  "persistent"
-#        instead of "synced" will result all writes getting skipped. Once
-#        the image does no longer contains this empty dir this patch can be
-#        reverted.
-/etc/cloud                              auto                    synced      none        none
+/etc/cloud                              auto                    persistent  none        none
```
anonymouse64, line 55: why was `action` set to `none` here? AFAICT, we want `transition` actually, as `action == none` means that we do nothing with the files already in the core20 snap, which means that files critical to cloud-init running such as `ds-identify.cfg` and friends are not installed on writable and thus cloud-init does not work properly when it should

mvo5: Probably a mistake/misunderstanding. Or wishful thinking in my crusade for empty /etc, i.e. that I hoped the defaults would not have to be stored there. Anyway, thanks for fixing this :)

PR #55, opened 2020-05-07
hooks: make systemd-modules-load depend on mounts at /usr/lib/{firmware,modules}

anonymouse64:
This is necessary because the initramfs will generate mounts in /etc/fstab for /usr/lib/{modules,firmware} from /run/mnt/kernel/{modules,firmware} respectively, and so before systemd-modules-load can load modules we need those things to be mounted.

This is opened as a draft until the required ubuntu-core-initramfs changes have landed.

Also to be clear, this is NOT for beta, so it can wait til next week.

### DIFF
--- a/hook-tests/110-add-modules-mount-deps-modules-load.test
+++ b/hook-tests/110-add-modules-mount-deps-modules-load.test
@@ -1,0 +1,4 @@
+#!/bin/sh -e
+
+grep "After=usr-lib-modules.mount" lib/systemd/system/systemd-modules-load.service.d/ubuntu-core-initramfs-mounts.conf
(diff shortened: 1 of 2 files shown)


anonymouse64:
Associated ubuntu-core-initramfs MP is open here: https://code.launchpad.net/~anonymouse67/ubuntu-core-initramfs/+git/ubuntu-core-initramfs/+merge/383583

anonymouse64:
This is not necessary anymore and is superceded by https://code.launchpad.net/~anonymouse67/ubuntu-core-initramfs/+git/ubuntu-core-initramfs/+merge/383614, which fixes the issue in a much simpler and more durable way

PR #53, opened 2020-05-07
Add secureboot-db

xnox:
Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

PR #56, opened 2020-05-07
Revert "Add secureboot-db"

xnox:
Reverts snapcore/core20#53


xnox:
@vorlonofportland talked with @mvo5 interactively, i think we want to ship secureboot-db data files somewhere, but only have gadget apply them, coupled with editions & validations2.0. But validations2.0 are not available yet. Thus reverting this for now, until there is coherent design, and once that's settled on, we then will be able to ship secureboot-db in the right snap (i.e. gadget, or core with service disabled, etc).

PR #58, opened 2020-05-15
hooks: enable check that we do not use "synced" dirs

mvo5:
Using "synced" in the writable-path something we do not want to
do in UC20. The user-experience is quirky, i.e. you can never
remove a file from a synced dir if that file is also available
via the underlying core snap. It will just reappear in the next
boot.


xnox:
Blocked on the other PR landing

PR #60, opened 2020-05-19
Copy-in launchpad's build-archive

xnox:
When building snap in launchpad, one can specify build archive
(Primary or PPA), and specify if release/security/updates/proposed
should be used for a given snap build.

Use that information during core20 snap build. This way, one can
rebuild core20 snap in personal PPAs, or rebuild it with Proposed
enabled, without any sourcecode/snapcraft.yaml modifications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ install:
 	tar -x --xattrs-include=* -f ../$(BASE) -C $(DESTDIR)
 	# ensure resolving works inside the chroot
 	cat /etc/resolv.conf > $(DESTDIR)/etc/resolv.conf
+	# copy-in launchpad's build archive
+	if grep -q ftpmaster.internal /etc/apt/sources.list; then \
+		cp /etc/apt/sources.list $(DESTDIR)/etc/apt/sources.list; \
+		cp /etc/apt/trusted.gpg $(DESTDIR)/etc/apt/ || true; \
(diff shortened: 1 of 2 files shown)


xnox:
Sample build:
https://launchpad.net/~xnox/+snap/xnox-core20/+build/961485

Note how the build was configured in my PPA with proposed enabled, and packages were in fact used from -proposed & my PPA.

Makefile
```diff
@@ -25,6 +25,12 @@ install:
 	tar -x --xattrs-include=* -f ../$(BASE) -C $(DESTDIR)
 	# ensure resolving works inside the chroot
 	cat /etc/resolv.conf > $(DESTDIR)/etc/resolv.conf
+	# copy-in launchpad's build archive
+	if grep -q ftpmaster.internal /etc/apt/sources.list; then \
+		cp /etc/apt/sources.list $(DESTDIR)/etc/apt/sources.list; \
+		cp /etc/apt/trusted.gpg $(DESTDIR)/etc/apt/ || true; \
```
mvo5, line 31: I guess we need the || true here because this file is not always available ?

xnox: Correct, it depends on target release and which archives are configured.

For example, Ubuntu Archive Bionic+ only ship /etc/apt/trusted.gpg.d/* snippets and don't have any /etc/apt/trusted.gpg

If PPAs are activated, one may have keys in /etc/apt/trusted.gpg in addition to those in /etc/apt/trusted.gpg.d/*

PR #62, opened 2020-05-21
Build consoleconf from git

xnox:
subiquity snap is built with snapcraft from git + archive deps.

core20 installs consoleconf from the archive, however core20 is a
snap, and it should build consoleconf from git too just like
subiquity.

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

I did two builds, of the current core20 and with this change applied.

$ sudo diff -ru old/ new/ 2>/dev/null | filterdiff -X '*.pyc' | diffstat
 manifest.yaml  |  407 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 snapcraft.yaml |   28 ++++++++++++
 2 files changed, 431 insertions(+), 4 deletions(-)

The net change is timestamps of .pyc files + manifest & snapcraft, otherwise the core20 snap is identical.

But now, instead of installing console-conf form the archive (which was built from the subiquity core/focal branch), during core20 snap build console-conf is built as a part from subiquity core/focal branch. This allows us to drop the round-trip to the archive, and simply push updates to core/focal whenever we need to update code in console-conf.

The sample core20 snap builds are:
- current https://launchpad.net/~xnox/+snap/xnox-core20/+build/967259
- this proposal https://launchpad.net/~xnox/+snap/xnox-core20-2/+build/966999

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -8,7 +8,35 @@ type: base
 build-base: core20
 
 parts:
+  consoleconf-deb:
+    plugin: nil
+    source: https://github.com/CanonicalLtd/subiquity.git
+    source-type: git
+    source-branch: core/focal
+    override-pull: |
+      snapcraftctl pull
+      # install build dependencies
+      export DEBIAN_FRONTEND=noninteractive
+      export DEBCONF_NONINTERACTIVE_SEEN=true
+      sudo -E apt-get build-dep -y ./
+    override-build: |
+      # unset the LD_FLAGS and LD_LIBRARY_PATH vars that snapcraft sets for us
+      # as those will point to the $SNAPCRAFT_STAGE which on re-builds will
+      # contain things like libc and friends that confuse the debian package
+      # build system
+      # TODO: should we unset $PATH to not include $SNAPCRAFT_STAGE too?
+      unset LD_FLAGS
+      unset LD_LIBRARY_PATH
+      # run the real build (but just build the binary package, and don't
+      # bother compressing it too much)
+      dpkg-buildpackage -b -Zgzip -zfast
+      cp ../console-conf_*.deb ../subiquitycore_*.deb $SNAPCRAFT_PART_INSTALL
+    stage:
+      - -console-conf_*.deb
+      - -subiquitycore_*.deb
   bootstrap:
+    after:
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -8,7 +8,35 @@ type: base
...
+      # unset the LD_FLAGS and LD_LIBRARY_PATH vars that snapcraft sets for us
+      # as those will point to the $SNAPCRAFT_STAGE which on re-builds will
+      # contain things like libc and friends that confuse the debian package
+      # build system
+      # TODO: should we unset $PATH to not include $SNAPCRAFT_STAGE too?
+      unset LD_FLAGS
+      unset LD_LIBRARY_PATH
+      # run the real build (but just build the binary package, and don't
+      # bother compressing it too much)
+      dpkg-buildpackage -b -Zgzip -zfast
```
vorlonofportland, line 32: I rather assumed if we would be pulling from git that we would be installing the contents directly from git rather than making a package and installing that.  What's the value in the packaging indirection here? /usr/share/doc/$pkg contents?  (/var/lib/dpkg is entirely absent from the snap, so it's not as if there's runtime value in it being added to that database.)  dependency resolution at install time?

xnox: The value is to ensure that during the switch to this everything is the same. Plus console-conf has a few things packaged using debhelper - i.e. systemd jobs, dist-packages Vs site-packages, etc.

Also note dependencies. When installing these two debs, extra debs are pulled from the archive - probert and urwid.

Later, after this lands we can switch from building deb to just using snapcraft python plugin. And stop having .deb packaging here. But care would need to happen to declare runtime debs, and install systemd units.

Note this is actually similar to snapd snapcraft which also builds deb, and then unpacks it.

xnox: These are installed whilst the base tarball still has dpkg, later all of that is cleaned out of the chroots prior to packing it as a base.

sil2100:
btw. are the Travis CI supposed to be failing? Or is it because of this PR?

PR #65, opened 2020-05-25
Do not attempt signing changes during snap build.

xnox:


xnox:
Hm, but we need to build the package on focal base =(

xnox:
@mvo5 thanks for this. Ideally, I'd want to be using focal to build things. Will look into launching lxd container inside travis with focal base.

PR #67, opened 2020-05-26
tests: use SNAPCRAFT_PRIME dir instead of hardcoding "prime"

mvo5:


xnox:
This is some black magic VOODOO

PR #71, opened 2020-05-27
tests: support test_etc-writable-symlinks.sh outside snapcraft

mvo5:
The test was relying on being run inside snapcraft. This is not
always the case so rework it to ensure it work inside snapcraft
and when run without snapcraft.


mvo5:
Closing this as #70 is already doing similar things

PR #73, opened 2020-06-16
hooks: fix broken symlink /etc/sysctl.conf.d/99-sysctl.conf

mvo5:
Core currently has a symlink /etc/sysctl.conf.d/99-sysctl.conf that
points to /etc/sysctl.conf. This file is removed by the core build
so removing this symlink is also needed.

Note that the sysctl.conf file is not writable on core and it
only contains commented out entries.

### DIFF
--- a/hooks/900-cleanup-etc-var.chroot
+++ b/hooks/900-cleanup-etc-var.chroot
@@ -23,7 +23,12 @@ rm /etc/pam.conf
 rm -rf /etc/profile.d
 
 rm /etc/rmt
+
+# sysctl.conf is not writable and everything in there is commented out
 rm /etc/sysctl.conf
+# 99-sysctl.conf is a symlink to /etc/sysctl.conf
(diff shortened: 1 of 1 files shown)


hooks/900-cleanup-etc-var.chroot
```diff
@@ -23,7 +23,12 @@ rm /etc/pam.conf
 rm -rf /etc/profile.d
 
 rm /etc/rmt
+
+# sysctl.conf is not writable and everything in there is commented out
 rm /etc/sysctl.conf
+# 99-sysctl.conf is a symlink to /etc/sysctl.conf
+rm /etc/sysctl.conf.d/99-sysctl.conf
```
pedronis, line 30: the dir is actually called "sysctl.d" as Dimitri mentioned

pedronis:
@mvo5 thanks, the core18/core variants also need the fix

PR #77, opened 2020-07-23
Drop console-conf hack, fixed in subiquity. core18 pr#148.

xnox:


xnox:
`After=core18.start-snapd.service core.start-snapd.service`

PR #79, opened 2020-07-30
Add secureboot-db package, try #2.

xnox:
Only execute it in install mode, and ensure this is done before snapd
is started.

The goal is to have dbxupdates applied before install is performed,
such that we seal against the latest dbx state possible.

### DIFF
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -80,3 +80,9 @@ find /tmp -name '*.deb' -print0 | xargs -0 apt install -y
 rm /tmp/*.deb
 
 apt autoremove -y
+
+case "$(dpkg --print-architecture)" in
+    amd64|i386)
+        apt install --no-install-recommends -y secureboot-db
+        ;;
+esac
--- a/static/usr/lib/systemd/system/core.start-snapd.service
+++ b/static/usr/lib/systemd/system/core.start-snapd.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=Start the snapd services from the snapd snap
 RequiresMountsFor=/run
+Wants=secureboot-db.service
+After=secureboot-db.service
 
 [Service]
 ExecStart=/usr/lib/core/run-snapd-from-snap start
--- a/static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf
+++ b/static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf
@@ -1,0 +1,2 @@
+[Unit]
+ConditionKernelCommandLine=snapd_recovery_mode=install


xnox:
Now, this still leaves the dbxupdates in the default keystore, and exposes sbkeysync in PATH. And if one executes this, one can brick themselves.

I think we have discussed to move the dbxupdate into non-standard keystore, or for example move it to snapd snap. But now I'm not sure where we want to ship the tooling and the dbxupdate. Advice needed here.

Othewise these changes work, and dbxupdate is applie don first boot _before_ snapd starts, and sealing is done against post-dbxupdate state, and one can boot after provisioning.

vorlonofportland:
On July 30, 2020 11:22:26 AM PDT, Dimitri John Ledkov <notifications@github.com> wrote:
>Only execute it in install mode, and ensure this is done before snapd
>is started.
>
>The goal is to have dbxupdates applied before install is performed,
>such that we seal against the latest dbx state possible.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/snapcore/core20/pull/79
>
>-- Commit Summary --
>
>  * Add secureboot-db package, try #2.
>
>-- File Changes --
>
>    M hooks/001-extra-packages.chroot (6)
>    M static/usr/lib/systemd/system/core.start-snapd.service (2)
>A
>static/usr/lib/systemd/system/secureboot-db.service.d/core-override.conf
>(2)
>
>-- Patch Links --
>
>https://github.com/snapcore/core20/pull/79.patch
>https://github.com/snapcore/core20/pull/79.diff
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/snapcore/core20/pull/79

Doesn't doing this in the core20 snap instead of the gadget snap introduce the possibility of a self-bricking image when the core snap is up to date but the gadget snap is shipping older assets?
-- 
Steve Langasek

xnox:
> Doesn't doing this in the core20 snap instead of the gadget snap introduce the possibility of a self-bricking image when the core snap is up to date but the gadget snap is shipping older assets?

Install mode is always executed from the recovery partition, which for the time being is immutable set of snaps.
If one builds an image which has "too new" dbxupdate that blacklists the gadgets or the kernel snaps artefacts, indeed the install mode boot will succeed, but subsequent boots in any mode will not. But I would call that a broken image, as the set of snaps that were used to construct the recovery partition are incompatible with each other.

Moving dbxupdate from core20 snap to the gadget doesn't solve the brickability, as you describe, as the gadget would still be able to ship dbxupdate that revokes kernel snap signatures.

Longer term, we have discussed moving dbxupdate data files to the snapd snap, if and when it grows ability to check if dbxupdate application will not brick the system; reseal to new values; apply dbxupdate; and on failure rollback resealing.

This PR only solves application of dbxupdates during (re)install only, to ensure our first seal is done against the best values possible.

Issue #72, opened 2020-06-12
move docker user/group to extrausers


anonymouse64:
Currently the docker user/group is defined in /etc/{passwd,group},
```
$ cat /snap/core20/current/etc/passwd | grep docker
docker:x:107:113:Reserved:/nonexistent:/bin/false
$ cat /snap/core20/current/etc/group | grep docker
docker:x:113:
```

This is problematic because it means that users can never be added to the docker group and thus use the docker snap as non-root on UC20. We have this bug pre-existing on UC16/UC18 as well, but apparently moving/transitioning users from /etc/passwd to /var/lib/extrausers/passwd is very dangerous, so for UC20 we should just move this user/group before the release so we don't have to do the complicated transition for UC20. 

What I tried which seemed to work is to replace the empty /var/lib/extrausers/{group,passwd} files in a built core20 snap with ones that contained the same definitions from /etc/{group,passwd} in the core20 snap right now. That unfortunately triggers https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1881588, but after working around that, I can then manually create a user that is in the docker group and this _almost_ works but still fails because the docker snap is too naive and isn't properly checking the /var/lib/extrausers/group file when it creates it's socket, but if it did then we would be good to go.

Also note that this is originally from a customer request about being able to use the docker snap as non-root on UC18, which is a different beast, but I think since this is easy enough to fix for core20, we should do that before UC20 is released.

CC @slimjim777 @ogra1

anonymouse64:
Also CC @tianon

tianon:
When using `extrausers`, doesn't `getent` inside the container return the "right" / smart result?  As in, if I do `getent group docker`, it should return the `extrausers` group data, right?

The Docker snap used to have a patch to explicitly check `extrausers`, but it was dropped because Docker upstream added a more flexible/general solution in that if they fail to find `docker` in `/etc/group`, they'll shell out to `getent`, so it sounds like maybe that's not working as expected? :disappointed:

xnox:
Blocked on releasing https://bugs.launchpad.net/snapd/+bug/1881588 to stable

ogra1:
be aware that the UID's/GID's must stay matching to not break on updates (there are user/group owned directories that are tied to the UID/GID on systems where docker has been used before).

there is a medium complex md5 check in the build scripts in livecd-rootfs that verifies if the actual /etc/group|passwd files have changed ...

anonymouse64:
@tianon 

[...]

PR #82, opened 2020-08-24
[RFC] hooks: mv docker user/group definition to extrausers

anonymouse64:
This will allow users to add themselves to the docker group. This in combination
with an upgraded getent which reads from extrausers will allow the docker snap
to be used with non-root users.

Fixes: https://github.com/snapcore/core20/issues/72

### DIFF
--- a/hooks/000-provide-uids-gids.chroot
+++ b/hooks/000-provide-uids-gids.chroot
@@ -115,7 +115,6 @@ systemd-network:x:104:109:systemd Network Management,,,:/run/systemd/netif:/bin/
 systemd-resolve:x:105:110:systemd Resolver,,,:/run/systemd/resolve:/bin/false
 systemd-bus-proxy:x:106:111:Reserved:/run/systemd:/bin/false
 _apt:x:117:65534:Reserved:/nonexistent:/bin/false
-docker:x:107:113:Reserved:/nonexistent:/bin/false
 syslog:x:108:114:Reserved:/home/syslog:/bin/false
 dnsmasq:x:109:65534:Reserved:/var/lib/misc:/bin/false
 tss:x:110:116:Reserved:/var/lib/tpm:/bin/false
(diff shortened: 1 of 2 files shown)


xnox:
Ideally we'd have `system-usernames` but we do not have that.

The group id / user-id must be preserved, and it is.

The only check is to ensure that it doesn't conflict with consoleconf, when lxd started to do similar things it tripped up first-boot experience into thinking that presence of extrausers at first-boot means that device is already managed, when it was not.

Obviously this is inert, unless docker snap is installed. And adding users to docker group must be done with care, as it opens up more privileges and is closer to root/sudo, then something confined.

anonymouse64:
we discussed most of this offline on mattermost, but just to be clear for the record here I will respond,

> The only check is to ensure that it doesn't conflict with consoleconf, when lxd started to do similar things it tripped up first-boot experience into thinking that presence of extrausers at first-boot means that device is already managed, when it was not.

we don't have such a test, but I'm not sure how to write such a test, since we would need to essentially use something like expect over a serial connection to the uc20 VM (granted virtual serial since it's a VM, but still) which I am afraid of for practical reasons. I did this test manually when building this snap and built a fresh image out of a core20 image from this PR and console-conf worked fine for me, so I think it's good, but we don't have proof of that beyond manual testing.

> Another thing to check, is what happens upon core20 snap refresh of existing beta instance with this inplace. Cause docker group will be dropped, but will it sync/get added to extra-users on writable parition, or not?

Existing beta users of uc20 will lose all definition of the docker group, however this is not a problem for folks using the docker snap because the status quo is that the docker group is defined in read-only /etc/group, so no user can ever be added to the docker group anyways and all actual users of docker on uc20 need to be sudo anyways. So if the docker group disappears, dockerd will log a message or something and happily fallback to listening on it's unix socket with the root group, which also effectively requires root/sudo to talk to docker, which is the same as the status quo, so no regression.

> And adding users to docker group must be done with care, as it opens up more privileges and is closer to root/sudo, then something confined.

yes of course, we do not add any users to the docker group automatically, and this will need to be done manually, even if the docker snap is preinstalled on the image, there will not be any users in the docker group by default (except I guess root by definition?)

xnox (approved):

PR #84, opened 2020-09-02
static/writable-paths: use transition instead of none for /etc/cloud

anonymouse64:
We still need files from the core20 snap to show up in writable for /etc/cloud,
otherwise cloud-init does not work correctly out-of-the-box. As such, use
transition instead of none, which will copy files from the core20 snap to
/etc/cloud on writable.

This fixes a bug introduced as part of #54 and will fix cloud-init with uc20 again.


pedronis (approved):
thank you

xnox (approved):

PR #87, opened 2020-09-23
Revert "hooks: mv docker user/group definition to extrausers"

xnox:
This reverts commit 224a6c62c5518eb8b26aa69b22063b5bd04e64d0.

### DIFF
--- a/hooks/990-ensure-uids-gids.chroot
+++ b/hooks/990-ensure-uids-gids.chroot
@@ -20,6 +20,18 @@ echo "Ensure gshadow file does not change"
 diff -u /etc/gshadow /etc/gshadow.orig
 rc=$?; [ "$rc" != "0" ] && MISMATCH=1
 
+echo "Ensure no docker"
+! grep -q docker /etc/passwd /etc/shadow /etc/group /etc/gshadow
+rc=$?; [ "$rc" != "0" ] && MISMATCH=1
+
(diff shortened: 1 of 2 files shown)


anonymouse64:
This is ok to go back to the status quo, but as per the meeting yesterday, we also want to drop the docker group from `/etc/{group,passwd}` entirely so there are no references to a docker group at all anywhere on a UC20 system, so while in the right direction I don't think this is sufficient

anonymouse64 (approved):
lgtm, thanks for adding the additional checks

hooks/990-ensure-uids-gids.chroot
```diff
@@ -20,20 +20,16 @@ echo "Ensure gshadow file does not change"
 diff -u /etc/gshadow /etc/gshadow.orig
 rc=$?; [ "$rc" != "0" ] && MISMATCH=1
 
-echo "Ensure extrausers passwd file does not change"
-diff -u /var/lib/extrausers/passwd /var/lib/extrausers/passwd.orig
+echo "Ensure no docker"
+! grep -q docker /etc/passwd /etc/shadown /etc/group /etc/gshadow
```
anonymouse64, line 24: mmm travis seems unhappy about this line

xnox: works now.

PR #95, opened 2020-11-16
Rpi kernel names

xnox:


vorlonofportland:
what are the specific device names that will change as a result of this PR? (from/to)

xnox:
> what are the specific device names that will change as a result of this PR? (from/to)

those that got configured by cloud-init will not change, as cloud-init emits "set-name" stanza.

wlan0 will remain wlan0.

enxb827ebc2edcd will become eth0, as it was on UC16/18 as UC16/18 booted with net.ifnames=0 which is no longer present on UC20.

There is no platform support to detect stable names without ACPI. I have filed about adding support for rpi stable platform names at https://github.com/systemd/systemd/issues/17625

This is now tested on classic and core, and works correctly.

vorlonofportland (approved):

PR #99, opened 2020-12-14
hooks: update xdg-settings to support subcommands

seb128:
Update to the current version of https://github.com/snapcore/snapd/blob/master/tests/lib/snaps/test-snapd-xdg-settings/bin/xdg-settings-wrapper which supports subcommands

Currently settings the default mailto handler from thunderbird fails because the installed wrapper doesn't handle subcommands, which was added in snapd with https://github.com/snapcore/snapd/pull/7129 but the core wrapper never got updated

Testcase
$ snap install thunderbird
$ xdg-settings set default-url-scheme-handler mailto thunderbird_thunderbird.desktop
$ snap run --shell thunderbird
$ xdg-settings get default-url-scheme-handler mailto
$ xdg-settings get default-url-scheme-handler mailto
NOT-THIS-SNAP.desktop
$ dbus-send --session --dest=io.snapcraft.Settings --type=method_call --print-reply /io/snapcraft/Settings io.snapcraft.Settings.GetSub string:'default-url-scheme-handler' string:'mailto'
method return time=1607963212.876053 sender=:1.387 -> destination=:1.388 serial=6 reply_serial=2
   string "thunderbird.desktop"

Checking /usr/bin/xdg-settings it's outdated, in the thunderbird environment with the current snapd tests wrapper

$ bash xdg-settings-wrapper get default-url-scheme-handler mailto
thunderbird.desktop

The update there is copying the current content from tests/lib/snaps/test-snapd-xdg-settings/bin/xdg-settings-wrapper so should be working

### DIFF
--- a/hooks/500-create-xdg.chroot
+++ b/hooks/500-create-xdg.chroot
@@ -31,24 +31,40 @@ EOF
 
 cat >/usr/bin/xdg-settings <<'EOF'
 #!/bin/bash
-#
 set -e
 set -o pipefail
-
-# we need to add a "cut -b4-" everyhwere because
-# `dbus-send --print-reply=literal` indents the output by 3 spaces
-# (hardcoded in dbus-print-message.c:print_iter)
-
-if [ "$1" = "get" ]; then
-   dbus-send --print-reply=literal --session --dest=io.snapcraft.Settings /io/snapcraft/Settings io.snapcraft.Settings.Get string:"$2" | cut -b4-
-elif [ "$1" = "check" ]; then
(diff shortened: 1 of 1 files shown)


anonymouse64 (approved):
thanks for this, would you mind opening a backport PR to at least core18, possibly also core? (I'll let foundations folks comment on whether this should be backported to core for UC16 since that will be EOL soonish)

seb128:
https://github.com/snapcore/core18/pull/175 , let me know if you can for 'core' also but it's desktop specific and probably not required there

sil2100 (approved):
Same as for core18, LGTM.

PR #101, opened 2021-04-20
ci: add a Github Actions port of the Travis workflow

jhenstridge:
This PR contains a port of the Travis CI workflow over to Github Actions.  As the travis-ci.org website has a big banner saying that it will soon be shut down, it seems prudent to move to an alternative.

I've made use of the `snapcore/action-build@v1` action in the workflow, which takes care of installing Snapcraft and running the build via LXD (so it is isolated from the Ubuntu install on the Github runner).  The workflow is set to run on pull requests and pushes to branches matching `master*`.  It won't run for this pull requests until the target branch contains the workflow though.

A sample run from my own repository can be seen here:

https://github.com/jhenstridge/core20/actions/runs/766038505

If this looks interesting, it should be pretty easy to copy for core, core18 and core22.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,3 +20,7 @@ jobs:
         run: |
           sudo unsquashfs -d prime ${{ steps.snapcraft.outputs.snap }}
           make test
+      - uses: actions/upload-artifact@v2
+        with:
+          name: snaps
+          path: ${{ steps.snapcraft.outputs.snap }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,22 @@
+name: Test
+on:
+  pull_request:
+  push:
+    branches: [ 'master*' ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install testing deps
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get update -q
+          sudo apt-get install -qy make squashfs-tools
+      - uses: snapcore/action-build@v1
+        id: snapcraft
+      - name: Test
+        run: |
+          sudo unsquashfs -d prime ${{ steps.snapcraft.outputs.snap }}
+          make test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,17 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Install testing deps
-        run: |
-          export DEBIAN_FRONTEND=noninteractive
-          sudo apt-get update -q
-          sudo apt-get install -qy make squashfs-tools
       - uses: snapcore/action-build@v1
         id: snapcraft
-      - name: Test
-        run: |
-          sudo unsquashfs -d prime ${{ steps.snapcraft.outputs.snap }}
-          make test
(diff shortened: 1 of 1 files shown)


xnox:
> This PR contains a port of the Travis CI workflow over to Github Actions. As the travis-ci.org website has a big banner saying that it will soon be shut down, it seems prudent to move to an alternative.
> 

It's not shutting down, it's moving from .org to .com domain. that's all. So it's not as drastic as all travis is going away.

> I've made use of the `snapcore/action-build@v1` action in the workflow, which takes care of installing Snapcraft and running the build via LXD (so it is isolated from the Ubuntu install on the Github runner). The workflow is set to run on pull requests and pushes to branches matching `master*`. It won't run for this pull requests until the target branch contains the workflow though.
> 
> A sample run from my own repository can be seen here:
> 
> https://github.com/jhenstridge/core20/actions/runs/766038505
> 

Running stuff in lxd with snapcraft is nice. And others seem to prefer github actions these days. So this does look good.

> If this looks interesting, it should be pretty easy to copy for core, core18 and core22.

Sure.

xnox:
hm, i am pondering why we need to run tests separately again. I thought all tests are executed during snapcraft build already..... Let me check that.

Also, is it at all possible to use remote launchpad build from github actions? Because travis has more architectures than github actions do - i.e. it has s390x ppc64el arm64, whereas github actions are amd64 only.

I wonder if we can switch to lxd & snapcraft builds in travis too, and keep travis for s390x / ppc64el / arm64.

Or like figure out how to do remote builds.

anonymouse64:
There are other projects that use snapcraft remote-build from within a GitHub Action like this, but then you have to deal with storing tokens, etc. I would consider that a nice to have some day and don't think that should block this PR necessarily. 

Also, other projects that do setup snapcraft remote-build from within a GitHub Action like this have suffered from the fact that sometimes snapcraft remote-build just hangs indefinitely due to a LP outage or some such thing and so they have to go in and manually cancel jobs and such. Maybe it's not such a big deal since this project only uses CI via GitHub for PR's and not for releases really, but still something to consider.

jhenstridge:
Previous communication from Travis said that the switch from travis-ci.org to .com will also involve converting from a free open source plan to a trial plan with a limited number of minutes.  They've said you can request more minutes when you run out, but the uncertainty is one of the things pushing people to alternatives.

As far as building a workflow around "snapcraft remote-build", one of the main problems is that it requires Launchpad user credentials.  While Github Actions allows you to create secrets, they are not made available to workflow runs triggered by pull requests (since they could run code from an untrusted source).  I guess it depends on what you want to get out of CI: have you run into many cases where everything was green on amd64, and then the snap failed to build later on?

xnox:
> Previous communication from Travis said that the switch from travis-ci.org to .com will also involve converting from a free open source plan to a trial plan with a limited number of minutes. They've said you can request more minutes when you run out, but the uncertainty is one of the things pushing people to alternatives.
> 

I see. Interesting.

> As far as building a workflow around "snapcraft remote-build", one of the main problems is that it requires Launchpad user credentials. While Github Actions allows you to create secrets, they are not made available to workflow runs triggered by pull requests (since they could run code from an untrusted source). I guess it depends on what you want to get out of CI: have you run into many cases where everything was green on amd64, and then the snap failed to build later on?

yes, there have been cases like that.

I'll see if we can create an account for launchpad just for this, to use remote build. And see if we can at least run that for the merged master branches. To at least get post-merge feedback.

xnox (approved):

PR #102, opened 2021-04-21
hooks: add mDNS hostname resolution support

jhenstridge:
This PR draft attempts to add optional support for mDNS hostname resolution via Avahi.  In order for it to be useful, Avahi will have to be available (either provided by the distro on classic systems, or via the avahi snap on Core).  If Avahi is not available, the plugin will fail quickly when it can't connect to `/run/avahi-daemon/socket`.

The AppArmor `<abstractions/nameservice>` policy fragment allows communication with this socket, so any snap plugging `network` already has the permissions it needs to communicate with Avahi's hostname resolution socket (which is heavily constrained compared to the D-Bus API).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ install:
 	# copy static files verbatim
 	/bin/cp -a static/* $(DESTDIR)
 	/bin/cp $(SNAPCRAFT_PART_INSTALL)/../../consoleconf-deb/install/*.deb $(DESTDIR)/tmp
+	/bin/cp $(SNAPCRAFT_PART_INSTALL)/../../avahi-daemon-equiv-deb/install/*.deb $(DESTDIR)/tmp
 	# customize
 	set -ex; for f in ./hooks/[0-9]*.chroot; do \
 		/bin/cp -a $$f $(DESTDIR)/tmp && \
--- a/hook-tests/013-add-mdns-hostname-resolution.test
+++ b/hook-tests/013-add-mdns-hostname-resolution.test
@@ -1,0 +1,26 @@
+#!/bin/sh -e
+
+# mdns support is enabled in nsswitch.conf
+grep -q "^hosts:.*mdns4_minimal" etc/nsswitch.conf
+
+# Avahi is not present in the base
+if [ -e usr/sbin/avahi-daemon ]; then
+    echo "The real avahi-daemon ended up in the image"
+    exit 1
+fi
+
+expected_plugins="$(find lib/ -name libnss_mdns4_minimal.so.2)"
+unexpected_plugins="$(find lib/ -name libnss_mdns.so.2 -o -name libnss_mdns_minimal.so.2 -o -name libnss_mdns4.so.2 -o -name libnss_mdns6.so.2 -o -name libnss_mdns6_minimal.so.2)"
+
+# There must be at least one copy of libnss_mdns4_minimal.so.2
+if [ -z "$expected_plugins" ]; then
+    echo "There must be a copy of libnss_mdns4_minimal.so.2 on the image"
+    exit 1
+fi
+
+# None of the other libnss-mdns plugins should be installed
+if [ -n "$unexpected_plugins" ]; then
+    echo "Found unexpected NSS plugins in image:"
+    echo "$unexpected_plugins"
+    exit 1
+fi
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -77,7 +77,7 @@ apt install --no-install-recommends -y \
     cryptsetup \
     gdbserver
 
-# Install self-built console-conf
+# Install self-built console-conf and fake avahi-daemon
 find /tmp -name '*.deb' -print0 | xargs -0 apt install -y
 rm /tmp/*.deb
 
--- a/hooks/012-add-foreign-libc6.chroot
+++ b/hooks/012-add-foreign-libc6.chroot
@@ -10,12 +10,6 @@ if [ "$(dpkg --print-architecture)" = "amd64" ]; then
 
     echo "I: Installing libc6:i386 in amd64 image"
     apt-get -y install libc6:i386
-
-    echo "I: Removing /var/lib/apt/lists/*"
-    find /var/lib/apt/lists/ -type f -print0 | xargs -0 rm -f
-
-    echo "I: Removing /var/cache/apt/*.bin"
-    rm -f /var/cache/apt/*.bin
 fi
 
 echo "I: Checking if we are arm64 and libc6:armhf should be installed"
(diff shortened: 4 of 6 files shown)


vorlonofportland:
My understanding is that this changes the behavior of resolution of the .local domain.  If so, that doesn't seem appropriate as a change for core20 (but it would for core22).

jhenstridge:
I created this in response to discussion on this forum thread:

https://forum.snapcraft.io/t/no-mdns-support-in-snaps-should-core-have-a-modified-nsswitch-conf/7603/20?u=jamesh

You are right that it will change the behaviour of resolution of `.local` domains, but only if Avahi is installed.  If the plugin cannot connect to the Avahi hostname resolution socket, it will report as UNAVAIL:

https://github.com/lathiat/nss-mdns/blob/99c2f1c10ff5b3373ce8c197add2563d66629f45/src/avahi.c#L120-L123

According to the nsswitch.conf man page, the default action for this status is to continue.  So on an Ubuntu Core device without Avahi installed, DNS based resolution will occur as before.  If you're building an Ubuntu Core device with Avahi included, it's not clear why you wouldn't expect Avahi to handle `.local` name resolution.

xnox:
@jhenstridge my understanding is that we already have support for this in core20

Did you try setting MulticastDNS=yes in `/etc/systemd/resolved.conf` ? Restart systemd-networked, and multicast dns resolution should start working then.

If that is not possible for some reason, i'd rather fix such that one can use systemd-resolved for mdns without avahi.

jhenstridge:
I'm primarily interested in getting mDNS resolution working by default on classic systems (e.g. have the Chromium snap be able to resolve .local hostnames on a default Ubuntu install). A solution that fixes things for one machine only isn't particularly interesting.

xnox:
> I'm primarily interested in getting mDNS resolution working by default on classic systems (e.g. have the Chromium snap be able to resolve .local hostnames on a default Ubuntu install). A solution that fixes things for one machine only isn't particularly interesting.

It is explicit and intentional that mDNS is disabled by default on both classic & core.

Note that changing or enabling mdns resolutions in core20 snap will not cause Chromium classic snap to use mDNS as it uses it's own resolvers based on the system resolvers directly.

jhenstridge:
mDNS hostname resolution seems to work out of the box on an Ubuntu desktop install.  The ubuntu-desktop meta-package recommends both avahi-daemon and libnss-mdns.

And are you sure about Chromium not ever using the system resolver? This README certainly seems to indicate that it uses the system resolver to ensure it can e.g. resolve `/etc/hosts` names:

https://chromium.googlesource.com/chromium/src/+/master/net/dns/README.md#Host-resolution-sources

There is also this forum thread and related bug that noted the change when Chromium switched from a deb to a snap:

https://forum.snapcraft.io/t/chromium-does-not-work-with-local-mdns-avahi-name-resolution/12483/12
https://bugs.launchpad.net/ubuntu/+source/chromium-browser/+bug/1838038

jhenstridge:
Snapd exposes the base snap's `nsswitch.conf` to strict confined snaps, even on classic systems.  This was done because the host system's nsswitch.conf could reference any collection of plugins, most of which won't be present in the sandbox: https://github.com/snapcore/snapd/pull/3650

Even if we were using the host system's config file, we'd need to have the plugins available within the sandbox for it to actually work.

As for the other concerns:

1. We've been installing avahi-daemon+libnss-mdns on desktop installs for a long time.  It's definitely there in precise, and I seem to remember it being there a lot longer than that. I do remember there was a privacy issue way back when we used to use Avahi's default of `publish-workstation=yes`, which caused it to publish a `._workstation._tcp` service.  That in turn allowed people to enumerate the hosts on the network.  None of that is necessary for this kind of hostname lookup though.

2. the libnss-mdns plugins do not (and can not) perform mDNS queries on their own.  When the plugin is queried, it is doing the equivalent of:
   ```
   $ echo 'RESOLVE-HOSTNAME-IPV4 scruffy.local' | nc -U /run/avahi-daemon/socket
   + 2 0 scruffy.local 192.168.0.121
   ```
   If there is no daemon listening on that socket, then no mDNS traffic is generated.  If a daemon is listening on the socket, the snap's AppArmor policy already allows it to issue such queries.

3. Any changes we make in future releases of Ubuntu to how mDNS hostname resolution is handled won't affect our existing user base.  We will have to deal with 20.04 LTS users right up to 2025.

xnox:
I think I am happy with doing minimal, ipv4 only, .local mdns resolution, for the 169.254.x.x range only, if there is server/daemon running on the localhost to respond to that.

I hope that nsswitch.conf looks sane with this change.

I would want to test this first by building & publishing this change on a branch, to see what the size impact is / diff, and if everyone is happy once this is in place.

I'm now also pondering if systemd-resolved doesn't implement "minimal, ipv4 only, .local, 169.254.x.x only resolutions" and if we can make that work in systemd-resolved upstream.

jhenstridge:
The only different line in the `nsswitch.conf` file after applying these changes is:

```
hosts:          files mdns4_minimal [NOTFOUND=return] dns
```

The snap I built locally is not noticeably larger than the version on the store:

```
james@scruffy:/var/lib/snapd/snaps$ ls -l core20_*
-rw------- 2 root root 64720896 Apr 20 14:53 core20_975.snap
-rw------- 1 root root 64692224 Apr 21 16:07 core20_x1.snap
```
The slight size reduction could be because my build didn't include the manifest, and a few library versions look slightly different.  It's small enough that it is likely to be lost in the noise.

As far as having systemd-resolved handle mDNS, it appears to be disabled by default on my 21.04 install:
```
$ resolvectl mdns
Global: no
Link 2 (enp5s0): no
Link 3 (lxdbr0): no
```
I imagine that if you enabled systemd-resolved's mDNS resolution code on a system without Avahi installed, it would correctly resolve `.local` hostnames with these changes applied.

tillkamppeter:
I have tested a [sample core20 Snap](https://forum.snapcraft.io/t/no-mdns-support-in-snaps-should-core-have-a-modified-nsswitch-conf/7603/34?u=till.kamppeter) with these changes and [everything is working](https://forum.snapcraft.io/t/no-mdns-support-in-snaps-should-core-have-a-modified-nsswitch-conf/7603/35?u=till.kamppeter).

jhenstridge:
Is there anything I could do to help move this PR forward?

On the implementation side, the main wart is the use of equivs-build to avoid including avahi-daemon and its dependencies in the snap. If there is a cleaner way of achieving that, I'd be happy to switch.

xnox (approved):

xnox:
> On the implementation side, the main wart is the use of equivs-build to avoid including avahi-daemon and its dependencies in the snap. If there is a cleaner way of achieving that, I'd be happy to switch.

if you really don't like equivs-build, one can do `apt download libnss-mdns` and then use `dpkg --force-depends -i ./libnss-mdns*.deb` to install it ignoring the depends.

jhenstridge:
I'd thought about doing something like that, but it leaves apt in a broken state, which could cause problems for the other hooks.  Perhaps equivs-build is the best way to handle this.

tillkamppeter:
@jhenstridge @xnox thank you very much! I have installed the new core20 from the edge channel now, rebuilt my CUPS Snap without the mDNS lookup workaround and it works. So the core20 does correct mDNS host name lookup now!

PR #103, opened 2021-05-18
hooks: add libglib2.0-bin for the "gdbus" utility

jhenstridge:
I'm proposing this in response to some discussion on IRC.  While writing snapd spread tests for some D-Bus related features, I found myself running into limitations on what I could do with the `dbus-send` utility.  The `gdbus` utility could handle those cases as it can represent arbitrary types when making method calls, but is not included in any of our base snaps. That means using it would require the tests use snaps published to the store, which increases the maintenance burden and makes updates more difficult. @anonymouse64 also noted that he'd run into similar limitations when writing tests.

I believe the `gdbus` utility would also be generally useful in testing D-Bus related faults on Ubuntu Core systems, so it would make sense to include for similar reasons to the `dbus-send` utility.

This PR adds the whole `libglib2.0-bin` package.  I don't expect it to make a large difference to the size of the snap since its dependencies are already included, but I'll check the result of CI. We could trim some of the other utilities from the package if you think that would be appropriate.

### DIFF
--- a/hook-tests/001-extra-packages.test
+++ b/hook-tests/001-extra-packages.test
@@ -68,6 +68,7 @@ libfdisk1:amd64
 libffi7:amd64
 libgcc-s1:amd64
 libgcrypt20:amd64
+libglib2.0-bin
 libgmp10:amd64
 libgnutls30:amd64
 libgpg-error0:amd64
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -61,6 +61,7 @@ apt install --no-install-recommends -y \
     less \
     dbus \
     dbus-user-session \
+    libglib2.0-bin \
     vim-tiny \
     bash-completion \
     netcat-openbsd \


anonymouse64:
lgtm, thanks for this

jhenstridge:
The snap spat out by CI is 65536 bytes larger than the core20 snap currently on the edge channel.  The new files are:

```
+squashfs-root/usr/bin/gapplication
+squashfs-root/usr/bin/gdbus
+squashfs-root/usr/bin/gio
+squashfs-root/usr/bin/gio-querymodules
+squashfs-root/usr/bin/glib-compile-schemas
+squashfs-root/usr/bin/gresource
+squashfs-root/usr/bin/gsettings
+squashfs-root/usr/share/bash-completion/completions/gapplication
+squashfs-root/usr/share/bash-completion/completions/gdbus
+squashfs-root/usr/share/bash-completion/completions/gio
+squashfs-root/usr/share/bash-completion/completions/gresource
+squashfs-root/usr/share/bash-completion/completions/gsettings
+squashfs-root/usr/share/doc/libglib2.0-bin
+squashfs-root/usr/share/doc/libglib2.0-bin/changelog.Debian.gz
+squashfs-root/usr/share/doc/libglib2.0-bin/copyright
+squashfs-root/usr/share/doc/libglib2.0-data
+squashfs-root/usr/share/doc/libglib2.0-data/changelog.Debian.gz
+squashfs-root/usr/share/doc/libglib2.0-data/copyright
+squashfs-root/usr/share/lintian/overrides/libglib2.0-bin
```

The `gio-querymodules` and `glib-compile-schemas` tools looked like they wouldn't be particularly useful, but it seems we're just adding symlinks to the tools already shipped in `$SNAP/lib/x86_64-linux-gnu/glib-2.0`.

xnox:
busctl would be smaller then both dbus-send & gdbus utility.

Is busctl from systemd not sufficient for you?

jhenstridge:
I hadn't played much with busctl, but it looks like it should fit the bill.  In my case, I had a D-Bus method call with an `(sa{sv})` argument (i.e. variant embedded in dict embedded in struct), and it looks like its syntax is expressive enough to marshal an argument like that:

    '(sa{sv})' unix-process 3 pid u "$pid" start-time t "$pid_start" uid u "$uid"

I think that covers pretty much everything I would have wanted to use gdbus for.

PR #105, opened 2021-07-27
hooks: enable systemd-time-wait-sync.service

mvo5:
The snapd team recently got bugreport about refresh issues that
look related to an incorrect time for devices without a battery
clock like the Raspberry Pi.

Upon inspection it looks like we do not enable the
systemd-time-wait-sync.service on core20 and this means that
we cannot ensure that snapd run only *after* the time is synced.
It seems right now we hit the race that timesyncd starts before
snapd but if it's slow at syncronizing snapd will start with the
wrong time.

Once we have systemd-time-wait-sync.service we can add:
```
After=time-sync.target
```
and that should eliminate this race.

However I would appreciate the (cirtical) review of the systemd maintainer(s).
It seems this feature is not without issues, e.g. https://github.com/systemd/systemd/issues/8683 or https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1937238


xnox (approved):

anonymouse64:
Note that this was reverted temporarily in #110 due to issues we found with fresh devices never finishing seeding in some specific environments due to issues with time synchronization

PR #108, opened 2021-08-16
Generate dpkg.yaml file

ilasc:
Proposed branch creates the dpkg.yaml file needed for the OSS compliance project (per discussions in design doc "LP101 - Minimal support for OSS compliance on selected snaps").


ilasc:
Tested by building the snap locally, contents of generated dpkg.yaml during build: https://pastebin.canonical.com/p/TNzr22pM8Y/

sil2100 (changes requested):
Same as for the core18 counterpart:

Ok, I think this looks decent from an initial look. So I'm +1 on the changes.

But I have one request: could you please add a simple test case for it into hook-tests/? It feels like something we could test, e.g. mocking the addition of a ppa, checking if the resulting yaml file has it. Mock a dpkg.list and checking if expected packages are present. Thanks!

stulluk (approved):

ilasc:
Following changes discussed for the Core Snap (https://github.com/snapcore/core/pull/121) now using `dpkg-query` to generate the packages section in the yaml file and removed package versions from the test.

hook-tests/600-no-debian.test
```diff
@@ -0,0 +1,268 @@
...
+  exit 1
+fi
+
+pkgs_start=$(( $(awk '/packages:/{ print NR; exit }' $FILE) +1))
+in_yaml_file_pkgs=$(cat <<EOF> in_yaml_file_pkgs.txt
+$(tail -n +$pkgs_start $FILE | awk -F= '{print $1}' | awk -F "- " '{print $NF }'|sort)
+EOF
+)
+
+expected_pkgs=$(sort cat <<EOF> expected_pkgs.txt
```
sil2100, line 33: Right now this generates an error `sort: cannot read: cat: No such file or directory` and generates an empty `expected_pkgs.txt`.
Should this simply be `expected_pkgs=$(sort <<EOF> expected_pkgs.txt` instead?

hook-tests/600-no-debian.test
```diff
@@ -0,0 +1,268 @@
...
+vim-tiny
+wpasupplicant
+xx
+zlib1g:amd64
+EOF
+)
+
+DIFF=$(comm in_yaml_file_pkgs.txt expected_pkgs.txt)
+echo $DIFF
+echo $DIFF1
```
sil2100, line 253: What is `$DIFF1`? Doesn't seem to be defined/used?

hook-tests/600-no-debian.test
```diff
@@ -0,0 +1,268 @@
...
+ucf
+udev
+util-linux
+vim-common
+vim-tiny
+wpasupplicant
+xx
+zlib1g:amd64
+EOF
+)
```
sil2100, line 249: Another thing regarding this: since the package list will most likely change, I don't think we need to test for such a huge set of packages. Might be actually enough for us to check if it's a valid yaml list of packages? Or maybe just check for a few key ones, if they're in the yaml.

hook-tests/600-no-debian.test
```diff
@@ -0,0 +1,268 @@
...
+else 
+  echo "Error! Packages section of dpkg.yaml not as expected."
+  echo $DIFF
+  echo "If that is intentional, please update the packages_section "
+  echo "content in the "
+  echo "test file: core18/hook-tests/602-create-dpkg-yaml.test."
+  exit 1
+fi
+
+# cleanup temp files used foe comparisons
```
sil2100, line 266:
```suggestion
# cleanup temp files used for comparisons 
```

sil2100 (changes requested):
This looks good, but the testing right now is a bit broken - it passes, but only because of issues with the tests. As mentioned in the inline comments, I think we should not be so detailed in the actual package contents testing - since the state is not something we can always guarantee. It's enough for us to check if the generated `dpkg.yaml` is a valid yaml file with the right PPA defined and (*some*) packages listed.

sil2100 (approved):
This looks much better, tests seem to look good as well.

sil2100:
Ok, same change approved by multiple people on the core18 repository - merging.

PR #109, opened 2021-08-19
hooks: add bootchart configuration

alfonsosanchezbeato:
Add configuration files so it is possible to get a bootchart on
boot. The systemd-bootchart.service will not be enabled by default,
but it will be possible to enable it from kernel command
line (systemd.wants=systemd-bootchart.service) if we want it to run on
first boot

### DIFF
--- a/hooks/024-configure-bootchart.chroot
+++ b/hooks/024-configure-bootchart.chroot
@@ -1,0 +1,76 @@
+#! /bin/sh -ex
+
+# Rewrite configuration and unit that came with the systemd-bootchart package
+
+cat << 'EOF' > /etc/systemd/bootchart.conf
+[Bootchart]
+Samples=36000
+Frequency=20
+Relative=yes
+Filter=no
+# Memory usage produces a bad overlay in the svg
+#PlotMemoryUsage=yes
+Cmdline=yes
+EOF
+
+cat << 'EOF' > /lib/systemd/system/systemd-bootchart.service
+[Unit]
+Description=Boot Process Profiler
+Documentation=man:systemd-bootchart.service(1) man:bootchart.conf(5)
+DefaultDependencies=no
+Conflicts=shutdown.target
+Before=shutdown.target
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p /run/log
+ExecStartPre=/lib/systemd/systemd-bootchart-prestart.sh
+ExecStart=/lib/systemd/systemd-bootchart -r
+KillSignal=SIGHUP
+ExecStopPost=/lib/systemd/systemd-bootchart-poststop.sh
+
+[Install]
+WantedBy=sysinit.target
+EOF
+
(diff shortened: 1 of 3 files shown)


stulluk (approved):
LGTM

stulluk (approved):

hooks/024-configure-bootchart.chroot
```diff
@@ -0,0 +1,74 @@
...
+
+cat << 'EOF' > /lib/systemd/system/stop-systemd-bootchart.service
+[Unit]
+Description=Unit to stop systemd-bootchart
+After=snapd.seeded.service
+Requisite=snapd.seeded.service
+
+[Service]
+Type=oneshot
+ExecStart=sh -c 'if /usr/bin/grep -q snapd_recovery_mode=run /proc/cmdline; then /usr/bin/systemctl stop systemd-bootchart.service; fi'
```
xnox, line 67: can you not use `ConditionKernelCommandLine=snapd_recovery_mode=run` ?
Then ExecStart would simply become `ExecStart=systemctl stop systemd-bootchard.service` no?
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionKernelCommandLine=

alfonsosanchezbeato: Good tip, I'll give that a try.

alfonsosanchezbeato: Changed and tested.

anonymouse64 (approved):
LGTM thanks for adding this

anonymouse64:
Also related question, at what point do we need to start filing PR's for this sort of improvement to core22 as well? I assume soonish?

stulluk:
> Also related question, at what point do we need to start filing PR's for this sort of improvement to core22 as well? I assume soonish?

I hope so :)

hooks/024-configure-bootchart.chroot
```diff
@@ -0,0 +1,75 @@
+#! /bin/sh -ex
+
+# Rewrite configuration and unit that came with the systemd-bootchart package
+
+cat << 'EOF' > /etc/systemd/bootchart.conf
+[Bootchart]
```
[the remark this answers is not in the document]
alfonsosanchezbeato: I have changed output to `/run/log/base` now. Saving directly to `ubuntu-save` is a bit more tricky because maybe the mountpoint is not available when `ExecStartPre` from `systemd-bootchart.service` is run, which is the point where the output folder would be created.

The nice thing with the change is that now we do not need systemd-bootchart-prestart.sh :)

hooks/024-configure-bootchart.chroot
```diff
@@ -0,0 +1,71 @@
+#! /bin/sh -ex
```
xnox, line 1: None of this should be a hook, but simply static files in static/ no?

I guess we can do that separately.

xnox (approved):

PR #110, opened 2021-09-01
hooks: revert PR#105 - it seems to 

mvo5:
It looks like adding 105 broke GCE tests. It's unclear why right now, the change works inside qemu. But to unbreak master we need to go back and revert this change.

I'm running a spread test with the services removed please do not merge before this is finished.


anonymouse64:
All the more reasons why it would be nice to have spread tests for core20 in github :smile:

xnox (approved):

PR #115, opened 2021-10-01
Start bootchart earlier

alfonsosanchezbeato:
Previously, we started systemd-bootchart by adding
systemd.wants=systemd-bootchart.service to the kernel command
line. This has as a side effect that the service is added to the
default target and is started later than desired. Instead, enable the
service by default and use a new kernel command line
parameter (core.bootchart) as condition to start it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,9 @@ doing a manual "make test" in the source tree.
 # Bootchart
 
 It is possible to enable bootcharts by adding
-`systemd.wants=systemd-bootchart.service` to the kernel command
+`core.bootchart` to the kernel command
 line. The sample collector will run until the system is seeded (it will
 stop when the `snapd.seeded.service` stops). The bootchart will be saved
 in the `ubuntu-save` partition, under `log/boot<N>/`, being `<N>` the
 boot number since bootcharts were enabled. If a chart has been collected
-by the initramfs, it will be also saved in that folder (this will happen
-if we have also added `rd.systemd.wants=systemd-bootchart.service` to
-the kernel command line).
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato:
See also https://github.com/snapcore/core-initrd/pull/46

mvo5 (approved):
Nice find! Thank you. 

One meta-question (kinda unrelated to this PR). The man-page tells me that:
```
Normally, systemd-bootchart is invoked by the kernel by
 passing init=/lib/systemd/systemd-bootchart on the kernel command line
```
which would give us even more correct results. Maybe worth adding to the README in a followup why this approach was not taken.

stulluk (approved):
LGTM

alfonsosanchezbeato:
> ```
> Normally, systemd-bootchart is invoked by the kernel by
>  passing init=/lib/systemd/systemd-bootchart on the kernel command line
> ```
> 
> which would give us even more correct results. Maybe worth adding to the README in a followup why this approach was not taken.

Rigth, ideally we want not only to start `systemd-bootchart` as PID 1, but also to ensure that it survives the switch root so we get only one bootchart that covers everything. However, when trying it as init process, I could not control how it was killed and no chart was being generated, so I decided to put it under systemd control. Also, `systemd-bootchart` code is not ready to survive the switch root (it would need to detect it and change its cwd).

So, as some modifications to `systemd-bootchart` were needed for that approach, we decided to go this route. But we might change that in the future (that would imply to carry some patches for systemd-bootchart though, hopefully temporarily).

I'll update the public docs with this information.

anonymouse64 (approved):
lgtm

PR #121, opened 2021-11-10
hooks: Add symlinks for bash completions of snap and snaps

valentindavid:
Completion for `snap` must be present in
`/usr/share/bash-completion/completions/` and it must point to the
file in the active Snapd.

Other completion will be added in
`/var/lib/snapd/share/bash-completion/completions/` by Snapd. To be
able to access those `XDG_DATA_DIRS` has to be set by Snapd's
`profile.d` file.

~~DRAFT: This commit only works with a modification of Snapd to store
completion files in writable `/var/lib/snapd` rather than in `/usr`.~~

This will require https://github.com/snapcore/snapd/pull/11038 to be merged to work. I will un-mark it as draft once the other PR is merged.

### DIFF
--- a/hooks/050-snap-symlink.chroot
+++ b/hooks/050-snap-symlink.chroot
@@ -2,3 +2,12 @@
 
 echo "Creating the snap binary symlink"
 ln -s /snap/snapd/current/usr/bin/snap /usr/bin/snap
+
+echo "Creating the symlink to snap's bash completion"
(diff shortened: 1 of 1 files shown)


valentindavid:
The changes required in snapd: https://github.com/snapcore/snapd/pull/11038

valentindavid:
Note that an example of application that works on ubuntu core and provides bash completion is docker.

PR #122, opened 2021-11-26
static/writable-paths: set safer options for tmpfs mounts

alfonsosanchezbeato:
Set safer options for mount points backed by tmpfs, so we make sure
that nosuid,nodev are set. The options are the default ones
recommended by systemd (see /usr/share/systemd/tmp.mount) and we were
actually using them already for /tmp in the initramfs.


static/etc/system-image/writable-paths
```diff
@@ -16,10 +16,10 @@
 # snap data
 /var/snap                               auto                    persistent  transition  none
 # generic
-/media                                  none                    temporary   none        defaults
-/mnt                                    none                    temporary   none        defaults
+/media                                  none                    temporary   none        mode=1777,strictatime,nosuid,nodev
```
mvo5, line 19: Do we need `mode=1777` for /mnt and /media ? Those are not really dirs that have the same semantic as /tmp, they are more dirs with a similar semantic as /run AIUI. Or is there soemthing I'm missing here?

alfonsosanchezbeato: Without this change, when I run `stat` on `/mnt` and `/media` it actually shows mode `1777` already, so I think that is the default mode. This just shows that more explicitly. So we can remove it I guess, I do not have a strong opinion either way.

alfonsosanchezbeato:
I've removed the change for `/var/lib/sudo` as anyway that folder can be used only by root.

anonymouse64 (approved):
lgtm thanks for this

Meulengracht (approved):
Cool, LGTM

PR #124, opened 2021-12-01
Add a finalrd script to move filesystems

valentindavid:
`/mnt/run/base` contains the snap file that is mounted on `/`. For
that reason it is not possible to unmount `/mnt/run/base` until we
have switched root to the shutdown ramfs.

At this point before we move `/sysroot/mnt/run/base` out of `/sysroot`
so that `systemd-shutdown` is able to unmount all filesystems.

`systemd-shutdown` allows hooks. But they are called right after
`systemd-shutdown`. That means we cannot run a hook after the root
switch and before `systemd-shutdown`. For that reason the
finalrd hook replaces `systemd-shutdown` in the ramfs by a wrapper.

This solution is not elegant. In the long term we should try to make
`systemd-shutdown` always move all sub mount points before trying to
unmount.

Also, there is still an error message when systemd tries to umount
`/mnt/run/base` before switching root. There does not seem to be
anyway to make `mnt-run-base.mount` perpetual.

### DIFF
--- a/static/usr/lib/systemd/systemd-shutdown.wrapper
+++ b/static/usr/lib/systemd/systemd-shutdown.wrapper
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# We need to manually move /run/mnt/data after shutdown switch root
+# so that systemd-shutdown can umount /oldroot in the shutdown initramfs.
+# systemd-shutdown has hooks, but they are run either before
+# switch, or after unmounts.
+# It would be better to make systemd-shutdown smarter and move
+# all found mounts before trying to unmount them.
+
+/bin/mount --make-private /oldroot/run
+for name in data ubuntu-seed; do
+  if /bin/mountpoint -q "/oldroot/run/mnt/${name}"; then
+    /bin/mkdir -p "/old-${name}"
+    /bin/mount --move "/oldroot/run/mnt/${name}" "/old-${name}"
+  fi
+done
+
+exec "${0}.real" "$@"
--- a/hooks/001-extra-packages.chroot
+++ b/hooks/001-extra-packages.chroot
@@ -74,7 +74,8 @@ apt install --no-install-recommends -y \
     wpasupplicant \
     cloud-init \
     dmsetup \
-    cryptsetup
+    cryptsetup \
+    initramfs-tools-core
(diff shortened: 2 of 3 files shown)


valentindavid:
Note there also snapcore/snapd#11119 to at least help `systemd-shutdown` to remount `/mnt/run/base` read-only. But there are still error messages.

static/usr/lib/systemd/systemd-shutdown.wrapper
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+# We need to manually move /run/mnt/data after shutdown switch root
+# so that systemd-shutdown can umount /oldroot in the shutdown initramfs.
+# systemd-shutdown has hooks, but they are run either before
+# switch, or after unmounts.
+# It would be better to make systemd-shutdown smarter and move
+# all found mounts before trying to unmount them.
```
xnox, line 8: somehow i would rather try to fix this in systemd-shutdown itself, instead of adding all of this into core20. Given that systemd-shutdown is C and can execute all of these things.

or for example, if some else can be done as a finalrd hook easier.

For example, can we have finalrd/core.finalrd hook that simply does `umount --lazy /run/mnt/*` ?

valentindavid: Yes, I will try to change `systemd-shutdown` to move filesystems in sub mount points. It should be an acceptable improvement for upstream.

The issue with `--lazy` is that it does not give you an error if it does not manage to unmount before shutdown. At least I am not aware of any way to check the active detached mounts. If only we had that, we could just mark `run-mnt-base.mount` as `LazyUmount=yes`, and verify it was really unmounted in a `system-shutdown` hook after `systemd-shutdown` has umounted `/sysroot`.

valentindavid:
We fixed it in systemd  in https://github.com/systemd/systemd/pull/25168 . So it will be fixed in uc24. And along this there is https://github.com/snapcore/core-base/pull/93 and https://github.com/snapcore/core-base/pull/83 which we could backport. On snapd side there was also https://github.com/snapcore/snapd/pull/11119
But this PR should not be merged. So I am closing this.

PR #127, opened 2022-01-12
Add a PR template that informs of core-base repository

Meulengracht:
This should help inform contributors when we do bugfixes for core20 that might also be relevant for core22+.


alfonsosanchezbeato (approved):
LGTM, thanks!

anonymouse64:
should also add something like this message to the README.md too

PR #129, opened 2022-01-26
Handle PPAs being served from ppa.launchpadcontent.net

cjwatson:
We now have a new HTTPS-capable domain for public PPAs, namely
ppa.launchpadcontent.net.  Adjust various bits of the build system to
accept that.


alfonsosanchezbeato (approved):
Thanks!

PR #130, opened 2022-02-02
static: remove perl dependency

kubiko:
remove perl dependency in base snap

Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>

for newer core versions.

### DIFF
--- a/static/usr/bin/core-sshd-host-keygen
+++ b/static/usr/bin/core-sshd-host-keygen
@@ -9,10 +9,8 @@ get_config_option() {
 	[ -f /etc/ssh/sshd_config ] || return
 
 	# TODO: actually only one '=' allowed after option
-	perl -lne '
-		s/[[:space:]]+/ /g; s/[[:space:]]+$//;
-		print if s/^[[:space:]]*'"$option"'[[:space:]=]+//i' \
-	   /etc/ssh/sshd_config
+	awk -v option=$option 'tolower($1) == tolower(option) {print $2}' \
+		/etc/ssh/sshd_config
 }
 
 host_keys_required() {


valentindavid (approved):
Approved.

valentindavid:
core-base PR snapcore/core-base#21

static/usr/bin/core-sshd-host-keygen
```diff
@@ -9,10 +9,8 @@ get_config_option() {
 	[ -f /etc/ssh/sshd_config ] || return
 
 	# TODO: actually only one '=' allowed after option
-	perl -lne '
-		s/[[:space:]]+/ /g; s/[[:space:]]+$//;
-		print if s/^[[:space:]]*'"$option"'[[:space:]=]+//i' \
-	   /etc/ssh/sshd_config
+	awk -v option=$option '$1 == option {print $2}' \
+        /etc/ssh/sshd_config
```
alfonsosanchezbeato, line 13: nitpick: you are mixing tabs (first line) with spaces (second line)

alfonsosanchezbeato (commented):
Thanks, looks good! Just a minor nitpick.

alfonsosanchezbeato (approved):

anonymouse64 (commented):
+0 from me because I don't understand perl, but the new awk bit looks correct so I think it's doing the right thing but because I don't know exactly what the original perl was doing I can't say for sure it's correct :smile:

kubiko:
One more change to address also comment from @valentindavid that sshd config is case insensitive.

kubiko:
@anonymouse64 that makes us two! :)
I just checked from the code what it actually expect to get back, rather than understand perl.

PR #131, opened 2022-02-02
No ssh key in install mode

kubiko:
Skip ssh key generation at install mode run

Not a big delay on fast devices, but saves some seconds on slow devices

### DIFF
--- a/static/usr/bin/core-sshd-host-keygen
+++ b/static/usr/bin/core-sshd-host-keygen
@@ -63,4 +63,7 @@ create_keys() {
 		"$hostkeys" /etc/ssh/ssh_host_ed25519_key -t ed25519
 }
 
-create_keys
+# create keys only if we are in run mode, there is no use to create keys in install mode
+if ! grep -q "snapd_recovery_mode=install" /proc/cmdline; then
+	create_keys
+fi


snapcraft.yaml
```diff
@@ -101,3 +101,77 @@ parts:
       snapcraftctl prime
       # ensure build-in tests are run
       cd ${SNAPCRAFT_PART_SRC} && make test TESTDIR=${SNAPCRAFT_PRIME}
+
```
anonymouse64, line 104: don't think all this is needed?

static/usr/bin/core-sshd-host-keygen
```diff
@@ -63,4 +61,7 @@ create_keys() {
 		"$hostkeys" /etc/ssh/ssh_host_ed25519_key -t ed25519
 }
 
-create_keys
+# create keys only if we are in run mode, there is no use to create keys in install mode
+if [ -n "$(grep "snapd_recovery_mode=install" /proc/cmdline)" ]; then
+  create_keys
+fi
```
anonymouse64, line 67: what about recovery mode ?

anonymouse64: that is, should we also skip generation for recovery mode as well? Ideally, recover mode would have the same host keys as from run mode by copying the files from ubuntu-data over to the tpmfs from the initramfs like we do for other things

snapcraft.yaml
kubiko: ups, yeah this is indeed not needed :)

static/usr/bin/core-sshd-host-keygen
kubiko: I was having same dilemma. But I was not sure what system is there in recovery mode.
Ultimately I opt for small step, as install mode is clear case and generated key is thrown away.

Also install mode is something which occurs on every first boot and in the factory, so extra time there is costly.
In comparison, recovery mode is rare in comparison, and more "please can you get me out of the trouble, I don't really care if it takes extra seconds, just help  me"

But happy to include recovery mode, as ultimately we end up in run mode and will genera they key there. You tell me preferred one :)

static/usr/bin/core-sshd-host-keygen
```diff
@@ -63,4 +63,7 @@ create_keys() {
 		"$hostkeys" /etc/ssh/ssh_host_ed25519_key -t ed25519
 }
 
-create_keys
+# create keys only if we are in run mode, there is no use to create keys in install mode
+if [ -n "$(grep "snapd_recovery_mode=install" /proc/cmdline)" ]; then
```
alfonsosanchezbeato, line 67: This should be, I think:
```suggestion
if ! grep -q "snapd_recovery_mode=install" /proc/cmdline; then
```

anonymouse64 (commented):
actually, thinking about this, we do have the `install-device` hook which runs/executes during install mode and may execute some factory specific logic, which could hypothetically use SSH to connect to some factory server and get some data, etc. 

If we don't create keys with this what happens? Is there a static default host SSH key on the device still or does the device then have no SSH key? Can you use SSH from the device to another one without having a host SSH key set/configured? I feel like the answer is yes, you can SSH out to another machine fine from the device being installed even if there are not host SSH keys generated, but another machine could not SSH to this device, which is expected/okay since there should never be a user account created in install mode to actually login via SSH.

PR #132, opened 2022-02-24
core: add pkcs11 dependencies

kubiko:
Add pkcs11 dependencies into core20 base snap

Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>

for newer core versions.

### DIFF
--- a/hooks/400-trim-pkcs-11.chroot
+++ b/hooks/400-trim-pkcs-11.chroot
@@ -1,0 +1,81 @@
+#!/bin/sh -ex
+
+mkdir -p /usr/share/p11-kit/modules
+echo "Creating p11-kit-client module configuration"
+echo "module: p11-kit-client.so" > \
+  /usr/share/p11-kit/modules/p11-kit-client.module
+
+# we should keep only following
+#   /usr/bin/p11-kit
+#   /usr/bin/p11tool
+#   /usr/bin/pkcs11-tool
+#   /usr/libexec
+#   /usr/lib/*/engines-*/*.so*
+#   /usr/lib/*/libopts.so*
+#   /usr/lib/*/libopensc.so*
+#   /usr/lib/*/libp11.so*
+#   /usr/lib/*/p11-kit-proxy.so*
+#   /usr/lib/*/pkcs11/*.so
+#   /usr/share/p11-kit
+
+echo "Cleaning excess pkcs#11 library"
+dpkg -r --force-depends libgnutls-dane0 libunbound8 libevent-2.1-7
+
+rm -rf \
+    /etc/opensc \
+    /usr/bin/cardos-tool \
+    /usr/bin/certtool \
+    /usr/bin/cryptoflex-tool \
+    /usr/bin/danetool \
+    /usr/bin/dnie-tool \
+    /usr/bin/egk-tool \
+    /usr/bin/eidenv \
+    /usr/bin/gids-tool \
+    /usr/bin/gnutls-cli \
+    /usr/bin/gnutls-cli-debug \
+    /usr/bin/gnutls-serv \
+    /usr/bin/goid-tool \
+    /usr/bin/iasecc-tool \
+    /usr/bin/netkey-tool \
+    /usr/bin/npa-tool \
+    /usr/bin/ocsptool \
+    /usr/bin/openpgp-tool \
+    /usr/bin/opensc-asn1 \
+    /usr/bin/opensc-explorer \
+    /usr/bin/opensc-notify \
+    /usr/bin/opensc-tool \
+    /usr/bin/piv-tool \
+    /usr/bin/pkcs11-register \
+    /usr/bin/pkcs15-crypt \
+    /usr/bin/pkcs15-init \
+    /usr/bin/pkcs15-tool \
+    /usr/bin/psktool \
+    /usr/bin/sc-hsm-tool \
+    /usr/bin/srptool \
+    /usr/bin/trust \
+    /usr/bin/westcos-tool \
+    /usr/lib/*/engines-1.1/pkcs11.la \
+    /usr/lib/*/libsmm-local.* \
+    /usr/lib/*/onepin-opensc-pkcs11.* \
+    /usr/lib/*/pkcs11-spy.* \
+    /usr/share/applications/org.opensc.notify.desktop \
+    /usr/share/bash-completion/completions/cardos-tool \
+    /usr/share/bash-completion/completions/cryptoflex-tool \
+    /usr/share/bash-completion/completions/dnie-tool \
+    /usr/share/bash-completion/completions/egk-tool \
+    /usr/share/bash-completion/completions/eidenv \
+    /usr/share/bash-completion/completions/gids-tool \
+    /usr/share/bash-completion/completions/iasecc-tool \
+    /usr/share/bash-completion/completions/netkey-tool \
+    /usr/share/bash-completion/completions/npa-tool \
+    /usr/share/bash-completion/completions/openpgp-tool \
+    /usr/share/bash-completion/completions/opensc-asn1 \
+    /usr/share/bash-completion/completions/opensc-explorer \
+    /usr/share/bash-completion/completions/opensc-notify \
+    /usr/share/bash-completion/completions/opensc-tool \
+    /usr/share/bash-completion/completions/piv-tool \
+    /usr/share/bash-completion/completions/pkcs15-crypt \
+    /usr/share/bash-completion/completions/pkcs15-init \
+    /usr/share/bash-completion/completions/pkcs15-tool \
+    /usr/share/bash-completion/completions/sc-hsm-tool \
+    /usr/share/bash-completion/completions/westcos-tool
--- a/hooks/400-trim-pkcs-11.chroot
+++ b/hooks/400-trim-pkcs-11.chroot
@@ -58,6 +58,7 @@ rm -rf \
     /usr/lib/*/libsmm-local.* \
     /usr/lib/*/onepin-opensc-pkcs11.* \
     /usr/lib/*/opensc-pkcs11.so \
+    /usr/lib/*/pkcs11/onepin-opensc-pkcs11.so \
     /usr/lib/*/pkcs11/opensc-pkcs11.so \
     /usr/lib/*/pkcs11-spy.* \
     /usr/share/applications/org.opensc.notify.desktop \
--- a/hooks/400-trim-pkcs-11.chroot
+++ b/hooks/400-trim-pkcs-11.chroot
@@ -61,6 +61,7 @@ rm -rf \
     /usr/lib/*/pkcs11/onepin-opensc-pkcs11.so \
     /usr/lib/*/pkcs11/opensc-pkcs11.so \
     /usr/lib/*/pkcs11-spy.* \
+    /usr/lib/*/pkcs11/pkcs11-spy.* \
     /usr/share/applications/org.opensc.notify.desktop \
     /usr/share/bash-completion/completions/cardos-tool \
     /usr/share/bash-completion/completions/cryptoflex-tool \
@@ -81,4 +82,5 @@ rm -rf \
     /usr/share/bash-completion/completions/pkcs15-init \
     /usr/share/bash-completion/completions/pkcs15-tool \
     /usr/share/bash-completion/completions/sc-hsm-tool \
-    /usr/share/bash-completion/completions/westcos-tool
+    /usr/share/bash-completion/completions/westcos-tool \
+    /usr/share/doc/opensc/README.md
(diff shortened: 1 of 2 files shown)


xnox (changes requested):
(duplicate)

xnox (changes requested):
1. Why do all classic & core users ever need these?
2. Why are you building from source instead of installing this packages from the archive?

xnox (changes requested):
Why are you building from source instead of installing this packages from the arhive?

And why do all classic & core users ever need these?

snapcraft.yaml
```diff
@@ -101,3 +101,84 @@ parts:
...
+    stage:
+      - etc
+      - usr/bin/p11tool
+      - usr/bin/p11-kit
+      - usr/libexec
+      - usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/libopts.so*
+      - usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/p11-kit-proxy.so*
+      - usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/pkcs11/p11-kit-client.so
+
+build-packages:
```
anonymouse64, line 171: this seems like it is meant to be a part of the `p11` part?

kubiko: do you mean 
`build-packages:`
no, you can define `build-packages:` in top level, rather for each part and repeat same dependency multiple times.
I assume this is more matter of preferences though...

anonymouse64: oh wow TIL top level `build-packages` is a thing I had no idea, always learning :smile:

snapcraft.yaml
```diff
@@ -101,3 +101,84 @@ parts:
...
+    plugin: autotools
+    source: https://git.launchpad.net/ubuntu/+source/opensc
+    source-type: git
+    source-depth: 1
+    source-branch: applied/ubuntu/focal
+    autotools-configure-parameters:
+      - --host ${SNAPCRAFT_ARCH_TRIPLET}
+      - --prefix=/usr
+      - --sysconfdir=/etc/opensc
+      - --with-pkcs11-provider=/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/pkcs11/p11-kit-client.so
```
valentindavid, line 115: I would prefer we use `p11-kit-proxy.so` and configure it to enable `p11-kit-client.so`. That will allow application to reconfigure with their own pkcs#11 module if they need without having to rebuild it.

snapcraft.yaml
```diff
@@ -101,3 +101,84 @@ parts:
...
+    plugin: autotools
+    source: https://git.launchpad.net/ubuntu/+source/libp11
+    source-type: git
+    source-depth: 1
+    source-branch: applied/ubuntu/focal
+    autotools-configure-parameters:
+      - --host ${SNAPCRAFT_ARCH_TRIPLET}
+      - --prefix=/usr
+      - --sysconfdir=/etc/opensc
+      - --with-pkcs11-module=/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}/pkcs11/p11-kit-client.so
```
valentindavid, line 138: Same here. `p11-kit-proxy.so` is actually the default on many distributions.

snapcraft.yaml
valentindavid: Interestingly, later in part `p11`, you configure p11-kit to load `p11-kit-client.so`. So that part is already done.

snapcraft.yaml
kubiko: I'm not sure there is good use case though for using `p11-kit-proxy.so` here.
- from within snap, we do not consider use case using any "actual" modules directly, it should always go through `p11-kit-client.so` to keep client<>service isolation.
- `p11-kit-client.so` does not handle cases to access two p11 server instances at the same time, as it takes config from env. So there is no actual way to connect to multiple p11-server instances using  `p11-kit-proxy.so` and configuring it to load `p11-kit-client.so` "multiple times".
- if we use  `p11-kit-proxy.so`, we will face added complexity, how to configure `/etc/pkcs11/modules/` for each snap. Should it be writable for each snap with pkcs11 plug? But snap should not be able to configure this globally, as we can have multiple pkcs11 slots, but given snap can have only some connected. So we would need to overlay it for each snap, while in 90% or more cases this will be just overkill for no value.
- using default `p11-kit-client.so` make it easy to consume for 90% cases, while snap can still override the default behaviour by using `--module` and load proxy instead.

I see the point for proxy on actual distributions, where you want to configure globally what modules to "chain" load by the proxy. But there you configure directly `opensc` and other modules. This won't be possible on core anyway. So it will not have desired benefit.

snapcraft.yaml
kubiko: For the proxy, please see my other comment.

kubiko: Good point for p11 part. Is the `/etc/pkcs11/modules/` only used by the `p11-kit-proxy.so`?
I thought it's also used by the `p11-tool` and reason I added it there.
With this config, `p11tool` also works without need to provide `--provider` argument

kubiko:
@xnox  that is actually a good idea! It'd solve multiple problems.
If we can change this in the archive and have it defaulted to p11-proxy module, then we can use stage packages rather then build from source and in that case, I'd be +1 with p11-proxy module as suggested by @valentindavid 

@xnox I assume we cannot do such a change to 20.04 archive though, can we?
Assuming we would add default proxy config file, so it still chain loads sc module

> > ad 2. PKCS11 tools and p11 library has defined "default" provider/module at the build time. Version in archive is defaulting to smart card (e.g. USB HSM), which is something not really used by embedded systems. Unfortunately there is config file to override this default behaviour, it really is only build time. And we need to set p11-client as default module/provider.
> 
> Is this something we can fix in the archive builds to improve classic Ubuntu too? For example should we make two builds in the archive? Or set the default provider to something else, that would be universally helpful? I.e. if it were set to `p11-kit-proxy.so` in the distro would that be useful on both regular laptops & on the embedded systems too?

xnox:
> @xnox that is actually a good idea! It'd solve multiple problems. If we can change this in the archive and have it defaulted to p11-proxy module, then we can use stage packages rather then build from source and in that case, I'd be +1 with p11-proxy module as suggested by @valentindavid
> 
> @xnox I assume we cannot do such a change to 20.04 archive though, can we? Assuming we would add default proxy config file, so it still chain loads sc module
> 

We just about have time to do it in 22.04 and we could attempt to change that in 20.04 too as long as the default module that p11-proxy loads by default is the one that used to be loaded on 20.04 too. Justification of greater compatibility for more things both on classic, core and across other distros is imho sane here and could win over SRU.


> > > ad 2. PKCS11 tools and p11 library has defined "default" provider/module at the build time. Version in archive is defaulting to smart card (e.g. USB HSM), which is something not really used by embedded systems. Unfortunately there is config file to override this default behaviour, it really is only build time. And we need to set p11-client as default module/provider.
> > 
> > 
> > Is this something we can fix in the archive builds to improve classic Ubuntu too? For example should we make two builds in the archive? Or set the default provider to something else, that would be universally helpful? I.e. if it were set to `p11-kit-proxy.so` in the distro would that be useful on both regular laptops & on the embedded systems too?

kubiko:
@xnox that sounds like a plan to me
And yes, my assumption was that if we move to p11-proxy, we would set default config the way that current behaviour stays the same.
Then we have parity between UC and classic and for that I'm +1 for p11-proxy.
Let's do it :)

> > @xnox that is actually a good idea! It'd solve multiple problems. If we can change this in the archive and have it defaulted to p11-proxy module, then we can use stage packages rather then build from source and in that case, I'd be +1 with p11-proxy module as suggested by @valentindavid
> > @xnox I assume we cannot do such a change to 20.04 archive though, can we? Assuming we would add default proxy config file, so it still chain loads sc module
> 
> We just about have time to do it in 22.04 and we could attempt to change that in 20.04 too as long as the default module that p11-proxy loads by default is the one that used to be loaded on 20.04 too. Justification of greater compatibility for more things both on classic, core and across other distros is imho sane here and could win over SRU.
> 
> Or for example we could add two flavour builds in 20.04, such that core20 can stage the proxy-flavour from the archive builds.
> 
> > > > ad 2. PKCS11 tools and p11 library has defined "default" provider/module at the build time. Version in archive is defaulting to smart card (e.g. USB HSM), which is something not really used by embedded systems. Unfortunately there is config file to override this default behaviour, it really is only build time. And we need to set p11-client as default module/provider.
> > > 
> > > 
> > > Is this something we can fix in the archive builds to improve classic Ubuntu too? For example should we make two builds in the archive? Or set the default provider to something else, that would be universally helpful? I.e. if it were set to `p11-kit-proxy.so` in the distro would that be useful on both regular laptops & on the embedded systems too?

kubiko:
@valentindavid @xnox I have changes default module to `p11-kit-proxy.so`
all tests are still working on my size, as we already have default config in place (`/etc/pkcs11/modules/p11-kit-client.module`)

@xnox content of this default config is `module: /usr/lib/<arch triplet>/pkcs11/p11-kit-client.so`
[See here when I create default config file](https://github.com/kubiko/core20/blob/pkcs11/snapcraft.yaml#L160)

kubiko:
@xnox where can I find default config parameters we use in the archive when building opensc and opensc-libp11?

xnox:
> @xnox where can I find default config parameters we use in the archive when building opensc and opensc-libp11?

pull-lp-source ? read debian/rules? or like find those packages in pad.lv/u/$pkg and read their build logs to see what the configury output says?

kubiko:
@xnox checking deb source package, there is already module configuration file, with correct module, so switching to the p11-proxy module will not change current behaviour (tested with USB HSM).
[change for opensc](https://paste.ubuntu.com/p/2sbVhQGP8b/)

As for libp11 does not have default module set, but when I'm testing with openssl which does load `libpkcs11.so` it does proceed loading proxy module.
So we could use that directly without change.

> > @xnox where can I find default config parameters we use in the archive when building opensc and opensc-libp11?
> 
> pull-lp-source ? read debian/rules? or like find those packages in pad.lv/u/$pkg and read their build logs to see what the configury output says?

kubiko:
@xnox @valentindavid can we please get some progress on this?

kubiko:
Force pushing, to redo PR by using hooks rather than snapcraft parts

xnox:
Still quite confused by this, given that opensc is being installed & removed; p11-kit proxy module is the default for libp11; and manual module config is shipped which is not otherwise present on classic. It would be nice to ship some default/better .module configs for p11-kit by default on classic too.

mvo5:
Also do we need this for core22 too?

xnox:
> Also do we need this for core22 too?

yes.

kubiko:
Rebase and push to remove dependency on ppa.

The agreement is to go ahead and use the pkcs11-tool from the archive as it is, to maintain current behaviour.
Most of the use cases are using openssl which uses libp11 which already defaults to the p11-proxy.

kubiko:
@mvo5 PR for core-base(core22) is also opened.

hooks/400-trim-pkcs-11.chroot
```diff
@@ -0,0 +1,84 @@
...
+    /usr/bin/srptool \
+    /usr/bin/trust \
+    /usr/bin/westcos-tool \
+    /usr/lib/*/engines-1.1/pkcs11.la \
+    /usr/lib/*/libsmm-local.* \
+    /usr/lib/*/onepin-opensc-pkcs11.* \
+    /usr/lib/*/opensc-pkcs11.so \
+    /usr/lib/*/pkcs11/onepin-opensc-pkcs11.so \
+    /usr/lib/*/pkcs11/opensc-pkcs11.so \
+    /usr/lib/*/pkcs11-spy.* \
```
valentindavid, line 63: Also remove the dead symlink `/usr/lib/*/pkcs11/pkcs11-spy.*`

valentindavid (approved):
This looks good. I have tested using nginx server in one snap using a token for pcscd from another snap. And that worked.

I have noticed on the snapd interface, on the plug side, `p11-kit-proxy.so` does not have read access to `/usr/share/p11-kit/modules` to find the `p11-kit-client.so`. It can be bind mounted. But I wonder if we should add it in the snapd pkcs11 interface.

There is also a dead link `/usr/share/doc/opensc/README.md`. I suppose this is an issue in the package. Maybe this should be reported.

Do not forget to open the PR in core-base.

hooks/400-trim-pkcs-11.chroot
kubiko: @valentindavid
 `/usr/share/doc/opensc/README.md` is link to `/usr/share/doc/opensc-pkcs11/README.md` which comes from `opensc-pkcs11` which we do not include. So I will trim that dead link as well
thank you for noticing those dead symbolic links!
yes core-base PR is also opened here: https://github.com/snapcore/core-base/pull/40

kubiko: @valentindavid as for `pkcs11` interface. I do not think this is needed, not for `p11-kit-client.so`, it's located at `/usr/lib/*/pkcs11/p11-kit-client.so`
Under `/usr/share/p11-kit/modules` is `opensc-pkcs11.module` which we do not intent to use from base snap. This would be typically included in gadget snap...

bugraaydogar:
@xnox Could you please prioritize the review of this pull request?

Thanks

xnox (approved):

PR #135, opened 2022-06-20
static: wait with snapd bootstrap after `time-set.target` is reached

mvo5:
We are investigating a case where a missing/broken RTC prevents
the initial seeding. On an empty/unseeded system the
`core.start-snapd.service` needs to wait for the initial (quick)
`time-set.target` too.

for newer core versions.


mvo5:
Now that https://github.com/snapcore/core20/pull/136 is merged and we understand the issues with the rtc we don't need this anymore.

PR #136, opened 2022-07-08
static: add fixrtc script and udev.rules

mvo5:
When the system has a RTC that is loaded as a module the time
may move backwards when the rtc module is loaded. See
https://forum.snapcraft.io/t/30391/15

This is a (racy) workaround to fix this.

See also the upstream bug:
https://github.com/systemd/systemd/issues/17737


static/usr/bin/fixrtc
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+NOW="$(date +'%s')"
+# XXX: also try /var/lib/systemd/timesync/clock ?
+MTIME_SELF="$(stat -L "/proc/self/exe" -c '%Y')"
```
mardy, line 5: This is brilliant!!

static/etc/udev/rules.d/90-fixrtc.rule
```diff
@@ -0,0 +1,2 @@
+# when rtc0 is added run fixrtc
+SUBSYSTEM=="rtc", KERNEL=="rtc0", RUN+= "/usr/bin/fixrtc"
```
mardy, line 2: Should this be
```suggestion
SUBSYSTEM=="rtc", KERNEL=="rtc*", RUN+= "/usr/bin/fixrtc"
```
in case there are more RTC devices?

mardy (commented):
LGTM, but let's see how testing goes first.

static/usr/bin/fixrtc
mardy: Actually, wait, I think you do not want `-L` here, otherwise you'll get the date of the script file.

static/etc/udev/rules.d/90-fixrtc.rule
mvo5: I think this makes sense, I will change it

static/usr/bin/fixrtc
mvo5: I think we need the `-L` - if this just stat()s /proc/self/exe it will read the 2000-01-01 (or 1970-01-01) date because the kernel time at this point is set to the rtc time which (at first boot) will be in the past. We need to move the time forward again and the baseline is the build/install time of the script on the filesystem. Or am I missing something?

static/etc/udev/rules.d/90-fixrtc.rule
```diff
@@ -0,0 +1,2 @@
+# when rtc* is added run fixrtc
+SUBSYSTEM=="rtc", KERNEL=="rtc*", RUN+= "/usr/bin/fixrtc"
```
mvo5, line 2: Fwiw, this should probably *not* live in /usr/bin but something like /usr/lib/core-fixrtc or something less user visible

static/usr/bin/fixrtc
```diff
@@ -0,0 +1,38 @@
...
+    LOG=/run/fixrtc-run.log
+fi
+exec >> "$LOG"
+exec 2>&1
+
+printf "%s\n" "Starting $(date)"
+TIMESYNC_CLOCK=/var/lib/systemd/timesync/clock
+SELF=$(readlink -f "$0")
+
+# get the mtime of this script
```
xnox, line 16: We should not use mtime of this script, and instead we should use `/usr/lib/clock-epoch` which is always provided in UC20+ initrd. It is the start time of when initrd was generated, and used by systemd to initialize clocks prior to loaded rtc modules causing havoc. No time should  ever be set to older than `/usr/lib/clock-epoch`.

static/etc/udev/rules.d/90-fixrtc.rule
mvo5: This is moved to /usr/lib/core/fixrtc now.

static/usr/bin/fixrtc
mvo5: Thanks, added this now.

xnox:
My last comment is about the name "fixrtc" - some people may incorrectly believe that it is the same as old initramfs-tools fixrtc feature, when it really isn't.

It is really is epoch-reset or like undo-modprobe-rtc-hctosys, as it's specific to dynamically loadaded rtc modules that wreck havoc to otherwise valid / already established time. This script is not specifically used to establish time, and normally is not executed at all, unlike what "fixrtc" in initramfs-tools used to do.

xnox:
Also very tempted to retest / resubmit this upstream and to all of our kernels:

```diff
commit 9431522c388c91c4d6dcd9f3a1ebd44d202a3b6e (HEAD -> prefix-gcc-suffix)
Author: Dimitri John Ledkov <dimitri.ledkov@canonical.com>
Date:   Thu Jul 28 14:01:57 2022 +0100

    drivers/rtc: do not perform rtc_hctosys upon modprobed rtc
    
    Do not perform RTC_HCTOSYS_DEVICE upon loading modularised RTC
    registration. Whilst preserving this behaviour for built-in RTC, and
    upon any resume from sleep.
    
    In general, distribution kernels configure and turn on
    CONFIG_RTC_HCTOSYS_DEVICE with usual set of RTC drivers built-in. This
    results in system clock being set prior to userspace being run, synced
    to hardware from NTP, and restore upon resume.
    
    There is a race between userspace & NTP acquired system time (from
    initrd for example, without RTC modules), and RTC kernel modules being
    loaded later upon udev trigger post pivot-root from the acutall root
    filesystem. At that point, it is desired to continue storing NTP time
    in RTC and restoring system time from RTC upon resume. But resetting
    system time from RTC upon module loading time is problematic, as it
    can be wrong relative to already userspace established system time via
    other means.
    
    Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@canonical.com>

diff --git a/drivers/rtc/class.c b/drivers/rtc/class.c
index 3c8eec2218df..02d9fb95b6f6 100644
--- a/drivers/rtc/class.c
+++ b/drivers/rtc/class.c
@@ -428,7 +428,7 @@ int __devm_rtc_register_device(struct module *owner, struct rtc_device *rtc)
                 dev_name(&rtc->dev));
 
 #ifdef CONFIG_RTC_HCTOSYS_DEVICE
-       if (!strcmp(dev_name(&rtc->dev), CONFIG_RTC_HCTOSYS_DEVICE))
+       if (!(rtc->owner) && !strcmp(dev_name(&rtc->dev), CONFIG_RTC_HCTOSYS_DEVICE))
                rtc_hctosys(rtc);
 #endif
```

mvo5:
The affected user reported the script fixes the issue. The log output:
```
+ date
+ printf %s
 Starting Sat Jan  1 00:00:53 UTC 2000
Starting Sat Jan  1 00:00:53 UTC 2000
+ TIMESYNC_CLOCK=/var/lib/systemd/timesync/clock
+ CLOCK_EPOCH=/var/lib/clock-epoch
+ readlink -f /usr/lib/core/fixrtc
+ SELF=/usr/lib/core/fixrtc
+ date +%s
+ NOW=946684853
+ stat -L /usr/lib/core/fixrtc -c %Y
+ MTIME_SELF=1658995608
+ MTIME_CLOCK_EPOCH=0
+ [ -e /var/lib/clock-epoch ]
+ MTIME_TIMESYNC_CLOCK=0
+ [ -e /var/lib/systemd/timesync/clock ]
+ stat -L /var/lib/systemd/timesync/clock -c %Y
+ MTIME_TIMESYNC_CLOCK=1650545679
+ echo 1658995608 0 1650545679
+ tr   

+ head -1
+ sort -rn
+ MTIME=1658995608
+ [ 946684853 -lt 1658995608 ]
+ echo Time needs updating because 946684853 < 1658995608
Time needs updating because 946684853 < 1658995608
+ date -s @1658995608
Thu Jul 28 08:06:48 UTC 2022
+ date
+ printf %s
 Starting Thu Jul 28 12:37:43 UTC 2022
Starting Thu Jul 28 12:37:43 UTC 2022
+ TIMESYNC_CLOCK=/var/lib/systemd/timesync/clock
+ CLOCK_EPOCH=/var/lib/clock-epoch
+ readlink -f /usr/lib/core/fixrtc
+ SELF=/usr/lib/core/fixrtc
+ date +%s
+ NOW=1659011863
+ stat -L /usr/lib/core/fixrtc -c %Y
+ MTIME_SELF=1658995608
+ MTIME_CLOCK_EPOCH=0
+ [ -e /var/lib/clock-epoch ]
+ MTIME_TIMESYNC_CLOCK=0
+ [ -e /var/lib/systemd/timesync/clock ]
+ stat -L /var/lib/systemd/timesync/clock -c %Y
+ MTIME_TIMESYNC_CLOCK=1650545679
+ echo 1658995608 0 1650545679
+ tr   

+ head -1
+ sort -rn
+ MTIME=1658995608
+ [ 1659011863 -lt 1658995608 ]
```
What is interesting from this log is that /usr/lib/clock-epoch is not there in the users test.

This should also be squash merged and forward ported to core22 - at least until a better fix is in the kernel or systemd.

xnox:
> What is interesting from this log is that /usr/lib/clock-epoch is not there in the users test.
> 

wait a minute, the clock-epoch file is in initrd only; and not available in the booted system, and we have the rtc modules in the booted system.

we should probably add clock-epoch during core snap builds too then.... or copy the one from initrd into runtime system, or like copy it into /run somewhere.

> This should also be squash merged and forward ported to core22 - at least until a better fix is in the kernel or systemd.

yeah.

static/usr/lib/core/fix-hctosys
```diff
@@ -0,0 +1,50 @@
...
+#else
+#    LOG=/run/fix-hctosys.log
+#fi
+#exec >> "$LOG"
+#exec 2>&1
+#printf "%s\n" "Starting $(date)"
+
+TIMESYNC_CLOCK=/var/lib/systemd/timesync/clock
+CLOCK_EPOCH=/var/lib/clock-epoch
+SELF=$(readlink -f "$0")
```
valentindavid, line 22: You could get rid of `SELF` by using `#!/bin/bash` and "stating" `${BASH_SOURCE}`. Just a suggestion. Not important.

static/usr/lib/core/fix-hctosys
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+# Get the mtime from timesyncd that is written every 60s
+MTIME_TIMESYNC_CLOCK=0
+if [ -e $TIMESYNC_CLOCK ]; then
+    MTIME_TIMESYNC_CLOCK="$(stat -L $TIMESYNC_CLOCK -c '%Y')"
+fi
+
+# find the highest MTIME from the two filesystem references
+MTIME=$(echo "$MTIME_SELF $MTIME_CLOCK_EPOCH $MTIME_TIMESYNC_CLOCK" | tr ' ' '\n' | sort -rn | head -1)
```
valentindavid, line 44: `stat` can look at multiple files. You could just put all in one line. Just ignore the errors.

MTIME="$(stat -L -c '%Y' "${TIMESYNC_CLOCK}  "${CLOCK_EPOCH}" "${SELF}" 2>/dev/null | sort -rn | head -n1)"

But this is just suggestion. Not important.

mvo5:
> > What is interesting from this log is that /usr/lib/clock-epoch is not there in the users test.
> 
> wait a minute, the clock-epoch file is in initrd only; and not available in the booted system, and we have the rtc modules in the booted system.
> 
> we should probably add clock-epoch during core snap builds too then.... or copy the one from initrd into runtime system, or like copy it into /run somewhere.

Thanks, indeed - I created https://github.com/snapcore/core20/pull/137 for this.

valentindavid (approved):
Looks fine.

mardy (approved):
LGTM, thanks!

PR #137, opened 2022-07-28
Makefile: create /usr/lib/clock-epoch

mvo5:
When there is no RTC systemd will use either it's own build date
or `/usr/lib/clock-epoch` to "seed" the initial time. So far we
did not create /usr/lib/clock-epoch but this commit adds it.

This should also get cherry picked for core22+

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ install:
                 fi && \
 		rm -f $(DESTDIR)/tmp/$$(basename $$f); \
 	done;
+	# see https://github.com/systemd/systemd/blob/v247/src/shared/clock-util.c#L145
+	touch $(DESTDIR)/usr/lib/clock-epoch
 
 	# only generate manifest and dpkg.yaml file for lp build
 	if [ -e /build/core20 ]; then \


valentindavid (approved):
Good.

alfonsosanchezbeato (approved):
LGTM, thanks!

FTR, systemd 250 introduced a way to periodically save the latest time on absence of NTP syncs (`SaveIntervalSec=`):
https://www.freedesktop.org/software/systemd/man/timesyncd.conf.html#SaveIntervalSec=
https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html#Files
Which might be useful to also configure on future UC versions.

alfonsosanchezbeato:
core-base PR: https://github.com/snapcore/core-base/pull/73

mvo5:
> LGTM, thanks!
> 
> FTR, systemd 250 introduced a way to periodically save the latest time on absence of NTP syncs (`SaveIntervalSec=`): https://www.freedesktop.org/software/systemd/man/timesyncd.conf.html#SaveIntervalSec= https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html#Files Which might be useful to also configure on future UC versions.

Indeed, iirc/afaik right now this is hardcoded to 60s (?)

PR #141, opened 2023-03-28
hooks: make sure that upgrade is never interactive

alfonsosanchezbeato:
In some cases builds have failed when configuration files in the new version to be upgraded are different from those in the build system.


alfonsosanchezbeato:
This is to fix core20 builds, which are currently broken: https://launchpadlibrarian.net/657015183/buildlog_snap_ubuntu_focal_amd64_core20_BUILDING.txt.gz

valentindavid (approved):

PR #142, opened 2023-04-11
Allow to customize faillock config

snowsky:
Adding faillock.conf file into writable-paths to be able to customize faillock config.
Will put up a PR to core-base after this is approved.


snowsky:
Hi @valentindavid, do you think this is good for approval? Thanks :)

Issue #143, opened 2023-04-18
$passwd not works properly since Core20 #1856 (arm64) and Core20 #1855 (armhf)


SoarHuang:
Original from [public LP #2012886](https://bugs.launchpad.net/snap-core20/+bug/2012886)
After auto refreshing core20 from `#1832` to `#1856`,
observe the new core20 `#1856` cannot run `$passwd` properly anymore.

```bash
iotuc@ubuntu:~$ passwd iotuc
Changing password for iotuc.
(current) UNIX password:
New password:
/var/cache/cracklib/cracklib_dict.pwd.gz: No such file or directory
BAD PASSWORD: The password fails the dictionary check - error loading dictionary
New password:
/var/cache/cracklib/cracklib_dict.pwd.gz: No such file or directory
BAD PASSWORD: The password fails the dictionary check - error loading dictionary
New password:
/var/cache/cracklib/cracklib_dict.pwd.gz: No such file or directory
BAD PASSWORD: The password fails the dictionary check - error loading dictionary

passwd: Have exhausted maximum number of retries for service
passwd: password unchanged

iotuc@ubuntu:~$
iotuc@ubuntu:~$ snap changes
ID Status Spawn Ready Summary
4 Done 4 days ago, at 04:42 UTC 4 days ago, at 04:44 UTC Auto-refresh snap "core20"

iotuc@ubuntu:~$
iotuc@ubuntu:~$ snap list
Name Version Rev Tracking Publisher Notes
core20 20230308 1856 latest/stable canonical✓ base
kernel 5.4.0-1034.45.1 126 latest/edge canonical✓ kernel
network-manager 1.22.10-14 743 20/stable canonical✓ -
snapd 2.58.3 18600 latest/stable canonical✓ snapd
gadget 20-3 71 latest/edge/… OEM/ODM gadget
iotuc@ubuntu:~$
```

SoarHuang:
Update more tests
If run with `$sudo`, it could change successfully, but it will show up 'error loading dictionary'

```bash
iotuc@ubuntu:~$ sudo passwd iotuc
New password:
/var/cache/cracklib/cracklib_dict.pwd.gz: No such file or directory
BAD PASSWORD: The password fails the dictionary check - error loading dictionary
Retype new password:
passwd: password updated successfully
```

SoarHuang:
Update more tests
If we add system-user via cloud-init (cloud.conf) in gadget snap,
we cannot login the UC20 system if setting 'expire: True',
[...]

PR #144, opened 2023-04-19
Add cracklib cache

valentindavid:
This is needed by pwquality. Fixes snapcore/core20#143.


mvo5 (approved):
Nice one!

hooks/601-clean-cracklib.chroot
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
```
mvo5, line 2: actually one quick comment - maybe a small readme why we remove most of the package, something like: `// we only need the cracklib_dict.pwd so that pamXXX works correctly, see LP:423423`

alfonsosanchezbeato (approved):
LGTM (but +1 in adding a comment explaining what we keep and why)

Meulengracht (approved):

hooks/601-clean-cracklib.chroot
valentindavid: Done.

PR #145, opened 2023-04-25
Use release instead of daily image since it was removed

valentindavid:


alfonsosanchezbeato (approved):
LGTM, thanks

valentindavid:
@alfonsosanchezbeato I added another commit to make the CI to pass. Is it still OK to merge?

xnox (approved):

xnox (changes requested):
Hm, that's a bit odd though. We rely on up to date daily images as a starting point.

@vorlonofportland @sil2100 is that intentional that ubuntu-base focal dailies got stopped & removed from cdimage? Is that correct that core20 snaps now must use out of date / frozen in time base to build?

vorlonofportland:
> Hm, that's a bit odd though. We rely on up to date daily images as a starting point.
> 
> @vorlonofportland @sil2100 is that intentional that ubuntu-base focal dailies got stopped & removed from cdimage? Is that correct that core20 snaps now must use out of date / frozen in time base to build?

No, this was an oversight when I pruned focal from the crontab post-20.04.6.  I'll add it back.

xnox:
http://cdimage.ubuntu.com/ubuntu-base/focal/daily/ are back, thanks a lot.

PR #149, opened 2023-06-21
bootchart: read mode from modeenv instead of kernel command line

valentindavid:
Also use a generator to enable it instead of disabling it with conditions.

Backport of snapcore/core-base#127


alfonsosanchezbeato (approved):
LGTM

PR #148, opened 2023-06-14
static: evolved fix-hctosys into rtc-init that also handles system time sync to rtc

ernestl:
Proposal to fix UC20 issue where system time is not synced to RTC.

Performance details: [SNAPDENG-7649](https://warthogs.atlassian.net/browse/SNAPDENG-7649?focusedCommentId=262768)

[SNAPDENG-7649]: https://warthogs.atlassian.net/browse/SNAPDENG-7649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #150, opened 2023-06-21
static/usr/lib: added more deterministic system time to rtc time sync activated by kernel cmdline

ernestl:
Followup proposal from [first](https://github.com/snapcore/core20/pull/148) to fix UC20 issue where system time is not synced to RTC.

Details and performance: 
- [SNAPDENG-7649 | Second proposal](https://warthogs.atlassian.net/browse/SNAPDENG-7649?focusedCommentId=262768)
- [SNAPDENG-7649 | Enhanced second proposal](https://warthogs.atlassian.net/browse/SNAPDENG-7649?focusedCommentId=265586)


static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,94 @@
...
+
+debug() {
+    if [ "$DEBUG_ENABLE" = "true" ]; then
+        log "$@"
+    fi
+}
+
+read_kernel_command_line() {
+    local cmdline="/proc/cmdline"
+    if [ -n "$SYSTEMD_VIRTUALIZATION" ] && [ ! "$SYSTEMD_VIRTUALIZATION" = "none" ]; then
```
valentindavid, line 25: I do not think we have to care about virtualization. I have not here of core20 being used with systemd-nspawn.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,94 @@
...
+    echo "$@" >> "/dev/kmsg"
+}
+
+debug() {
+    if [ "$DEBUG_ENABLE" = "true" ]; then
+        log "$@"
+    fi
+}
+
+read_kernel_command_line() {
```
valentindavid, line 23: There is `get-arg` that I have added in another PR that you can use.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,94 @@
...
+    _RET=$(cat $cmdline)
+    debug "Using command line $cmdline: _RET"
+    return 0
+}
+
+search_command_line() {
+    param=$1
+    if read_kernel_command_line; then
+        cmdline=$_RET
+        for cmd in $cmdline; do
```
valentindavid, line 38: This is not how cmdline is parsed.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,94 @@
...
+                log "Enable failed: cannot create directory $link_path"
+            fi
+        fi
+    else
+        log "Kernel command line requests disable"
+
+        if [ ! -L "$link_file" ]; then
+            log "Already disabled, no change needed"
+        else
+            if rm -f "$link_file"; then
```
valentindavid, line 82: I do not think you should remove anything. The issue is that if several generators were to enable for different reason, they should not fight for it. Generators should only enable things. Not disable.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
valentindavid: `/usr/libexec/core/get-arg`

static/usr/lib/udev/rules.d/90-rtc-sys-time-init.rules
```diff
@@ -0,0 +1,3 @@
+# When hctosys device rtc* is added create device unit dev-rtc*.
+# This is used by rtc-sys-time-init.service.
+ACTION=="add", SUBSYSTEM=="rtc", KERNEL=="rtc*", ATTR{hctosys}=="1", TAG+="systemd"
```
valentindavid, line 3: I think there is something wrong here. We have 2 features. One to rollback wrong update from `RTC_HCTOSYS_DEVICE`. This has to be always enabled. The other is to emulate hctosys for some rtc devices not configurated as `RTC_HCTOSYS_DEVICE`. This has to be enabled conditionally, and it should not be expected `/dev/rtc` (because not `ATTR{hctosys}`).

static/usr/lib/systemd/system/sync-system-to-rtc.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Synchronize system time with RTC time
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/core/sync-system-to-rtc
```
valentindavid, line 6: Maybe this script is missing.

static/usr/lib/systemd/system/rtc-sys-time-init.service
```diff
@@ -0,0 +1,29 @@
...
+# RTC /dev/rtcx to /dev/rtc and on udev rule 90-rtc-sys-time-init.rules to create 
+# device unit dev-rtc.device when the hctosys RTC device is ready.
+[Unit]
+Description=Correct system time and sync to RTC time
+SourcePath=/proc/cmdline
+DefaultDependencies=no
+After=dev-rtc.device
+Requires=dev-rtc.device
+Before=systemd-timesyncd.service
+Conflicts=shutdown.target
```
valentindavid, line 24: Missing `Before=shutdown.target`. Otherwise this does not do much.

static/usr/lib/systemd/system/rtc-sys-time-init.service
```diff
@@ -0,0 +1,29 @@
...
+ConditionKernelCommandLine=rtc_sys_time_init
```
valentindavid, line 25: Do we need this if we have a generator?

static/usr/lib/systemd/system/rtc-sys-time-init.service
```diff
@@ -0,0 +1,29 @@
...
+#  b.) Set system time to valid RTC time when kernel configuration hctosys is enabled
+#      with kernel configuration CONFIG_RTC_HCTOSYS and hctosys device is specified 
+#      with kernel configuration CONFIG_RTC_HCTOSYS_DEVICE.
+#
+# This service relies on 50-udev-default.rules to create symlink from hctosys specified 
+# RTC /dev/rtcx to /dev/rtc and on udev rule 90-rtc-sys-time-init.rules to create 
+# device unit dev-rtc.device when the hctosys RTC device is ready.
+[Unit]
+Description=Correct system time and sync to RTC time
+SourcePath=/proc/cmdline
```
valentindavid, line 19: This is only for generated files. This is not generated.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
ernestl: I removed this idea from the script

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
ernestl: Thanks, back-ported this into my code to enable testing.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
ernestl: Understood, removed link removal.

static/usr/lib/udev/rules.d/90-rtc-sys-time-init.rules
ernestl: Thanks for pointing this out. Indeed hctosys=1 only happens when driver actually sets system time, which is not the case in older kernel with the issue! Now addressed.

static/usr/lib/systemd/system/rtc-sys-time-init.service
ernestl: Thanks, added.

static/usr/lib/systemd/system/rtc-sys-time-init.service
ernestl: Good point, I thought it might still be useful, but in practice cmdline should not change. Removed.

static/usr/lib/systemd/system/rtc-sys-time-init.service
ernestl: No we do not, I removed it

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,54 @@
...
+
+# Look for kernel command line parameter rtc_sys_time_init
+if ! dev_node=$(/usr/libexec/core/get-arg "rtc_sys_time_init"); then
+    debug "Kernel command line does not contain parameter rtc_sys_time_init"
+    exit 0
+fi
+
+# Allow value /dev/rtc and /dev/rtc0 to /dev/rtc9 (Assume no more than 9 RTCs)
+# TODO: Can the order of enumeration be guaranteed/controlled?
+if ! expr "$dev_node" : '^/dev/rtc[0-9]\{0,1\}$' >/dev/null; then
```
valentindavid, line 29: I would not verify this in the generator. Otherwise it will go unnoticed if it does not work. Instead we want the service later to fail or timeout.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
```diff
@@ -0,0 +1,54 @@
...
+    log "Kernel command line parameter rtc_sys_time_init=$dev_node is invalid"
+    exit 1
+fi
+
+dev_unit=$(echo "$dev_node" | sed 's/\/dev\/rtc/dev-rtc/')
```
valentindavid, line 34: Can you use `systemd-escape` instead?

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,105 @@
...
+
+
+
+
+
+
+
+
+
+
```
valentindavid, line 105: Why so many empty lines?

static/usr/lib/core/fix-hctosys
```diff
@@ -17,6 +17,9 @@
 #exec 2>&1
 #printf "%s\n" "Starting $(date)"
 
+DEV_PATH="$1"
```
valentindavid, line 20: Maybe we do not need to pass as parameter.

Though it would be nice to check whether the device comes from a module. Because in that case we do not need to fix anything.

That said, we can also use an `IMPORT{program}` with the following script to do some filtering the udev rule instead.

```
#!/bin/sh
set -eu

device=/sys/${DEVPATH}

while true; do
    if [ -d "${device}/driver" ]; then
        if [ -L "${device}/driver/module" ]; then
            echo FROM_MODULE="$(basename "$(realpath "${device}/driver/module")")"
        fi
        break
    elif [ -d "${device}/device" ]; then
        device="${device}/device"
    else
        break
    fi
done
```

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,105 @@
...
+#     2) Correct driver/module system time sync to invalid time (earlier than most recent time stamp)
+#          - Built-in/loadable RTC driver with HCTOSYS enabled may set invalid time due to uninitialized,
+#            invalid time e.g. caused by bad battery
+#
+# If the RTC time is not valid, or cannot be set, forward the system time to most recent timestamp
+
+# Expects single argument that is service instance name which is
+# based on the associated RTC device unit name in the form dev-rtc*
+dev_unit="$1"
+
```
valentindavid, line 17: I feel like we should look at hctosys in the devpath, if it set to 1, we do not have to do anything here. That way we know we do not have conflict with the other script.

static/usr/lib/core/rtc-sys-time-init
ernestl: Accident, fixed

static/usr/lib/core/rtc-sys-time-init
valentindavid: Ah but the other script is not called because of `IMPORT{cmdline}="rtc_sys_time_init", ENV{rtc_sys_time_init}!="?*"`.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,78 @@
...
+    fi
+fi
+
+# Check if system time was successfully synced to RTC time
+if [ "$sys_time_set_res" -ne 0 ]; then
+    # Move system time forward if required
+    now_stamp="$(date +'%s')"
+    if [ "$now_stamp" -lt "$most_recent_stamp" ]; then
+        echo "Moving system time forward to most recent timestamp..."
+        if date -s @"$most_recent_stamp" >/dev/null 2>&1; then
```
valentindavid, line 68: This part should happen only when `hctosys` is 1 in sysfs and the rtc comes from a module.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,78 @@
...
+    sys_time_set_res=1
+    echo "Error: cannot read RTC $dev_node"
+else
+    rtc_stamp=$(date -d "$rtc_stamp" +%s)
+    rtc_date=$(date -d @"$rtc_stamp")
+    echo "Successfully read RTC $dev_node time: $rtc_date"
+
+    if [ "$rtc_stamp" -gt "$most_recent_stamp" ]; then
+        echo "RTC time is valid, syncing system time..."
+        if hwclock -s --noadjfile -u -f "$dev_node" 2>/dev/null; then
```
valentindavid, line 50: This should happen only when `hctosys` is 0.

static/usr/lib/core/fix-hctosys
valentindavid: Do not worry about this comment. This is the least urgent.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
ernestl: Changed it to only issue warning, for extra visibility.

static/usr/lib/core/fix-hctosys
ernestl: Moved away from RUN to SYSTEMD_WANTS based approach using same service template as the generator. To my understanding this limits ability to "pass arguments" to the service in a simple way.

static/usr/lib/core/rtc-sys-time-init
ernestl: Yes, the generator and udev triggers are mutually exclusive. Did incorporate looking for `hctosys` value in this script to prevent unnecessary re-sync for system time to RTC if driver already done it, as per your other suggestion.

static/usr/lib/core/rtc-sys-time-init
ernestl: Great suggestion, thanks. Done.

static/usr/lib/systemd/system-generators/rtc-sys-time-init-generator
ernestl: Nice, thanks!

static/usr/lib/core/rtc-sys-time-init
ernestl: Changed to happen when `hctosys` is 1 or if `hctosys` is 0, but RTC time invalid or setting failed.

static/usr/lib/systemd/system/rtc-sys-time-init@.service
```diff
@@ -0,0 +1,28 @@
...
+DefaultDependencies=no
+After=%i.device
+Requires=%i.device
+Before=systemd-timesyncd.service
+Conflicts=shutdown.target
+Before=shutdown.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/core/rtc-sys-time-init %I
```
valentindavid, line 28: I would use `/%I` here and remove the prefixing in the script.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,112 @@
...
+        journal error "Error: cannot read RTC $dev_node"
+    else
+        rtc_stamp=$(date -d "$rtc_stamp" +%s)
+        rtc_date=$(date -d @"$rtc_stamp")
+        echo "Successfully read RTC time: $rtc_date"
+
+        if [ "$rtc_stamp" -gt "$most_recent_stamp" ]; then
+            echo "RTC time is valid, syncing system time..."
+            if hwclock -s --noadjfile -u -f "$dev_node" 2>/dev/null; then
+                sys_time_set_res=0
```
valentindavid, line 84: I would exit here to make it easier to read.

static/usr/lib/udev/rules.d/90-rtc-sys-time-init.rules
```diff
@@ -0,0 +1,12 @@
+# When device rtc* is added
+# (1) and the RTC module have successfully set system time from RTC (indicated with hctosys attribute)
+# (2) and the kernel command line does not contain parameter rtc_sys_time_init that will cause
+#     systemd generator rtc-sys-time-init-generator to instantiate rtc-sys-time-init@.service
+# instantiate rtc-sys-time-init@.service for the device node
+IMPORT{cmdline}="rtc_sys_time_init"
+ACTION=="add", SUBSYSTEM=="rtc", KERNEL=="rtc*", ATTR{hctosys}=="1", ENV{rtc_sys_time_init}!="?*", TAG+="systemd", PROGRAM="/usr/bin/systemd-escape --path %N", ENV{SYSTEMD_WANTS}+="rtc-sys-time-init@%c.service"
```
valentindavid, line 7: I think we are missing some `IMPORT{cmdline}` to read `rtc_sys_time_init`.

valentindavid: Ah it is before. Though I would have but it after `ATTR{hctosys}=="1"`, but I think it is fine like that.

valentindavid (commented):
I like what I see now. I will do some testing and likely approve. Thank you.

static/usr/lib/udev/rules.d/90-rtc-sys-time-init.rules
ernestl: The IMPORT only becomes available to rules after the rule that imports. I have read this, and tested this, having it in the same line does not work.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,112 @@
...
+    if [ "$now_stamp" -lt "$most_recent_stamp" ]; then
+        echo "Moving system time forward to most recent timestamp..."
+        if date -s @"$most_recent_stamp" >/dev/null 2>&1; then
+            echo "Successfully moved system time forward "
+        else
+            sys_time_set_res=1
+            journal error "Error: cannot forward system time"
+        fi
+    else
+        echo "System time is ahead of most recent timestamp, skipping update"
```
valentindavid, line 108: I would not say "skipping update". This one is a fixup. I would call say "skipping fixup".

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,112 @@
...
+clock_epoch_date=$(date -d @"$clock_epoch_stamp")
+journal debug "core/systemd build timetamp: $clock_epoch_date"
+
+# Get systemd-timesyncd timestamp file initialized to systemd build time and updated periodically
+# (default 60s) and at shutdown
+timesync_clock_stamp=0
+if [ -e "/var/lib/systemd/timesync/clock" ]; then
+    timesync_clock_stamp="$(stat -L "/var/lib/systemd/timesync/clock" -c '%Y')"
+else
+    journal warning "Warning: cannot stat /var/lib/systemd/timesync/clock"
```
valentindavid, line 59: After we will have removed the broken hack in `901-cleanup-timesyncd.chroot`, this will still happen on the first boot. In order not to become a red herring, we should say that this may be expected on first boot in the warning message.

valentindavid (approved):
Looks good. There are few comments I left.

Could you also rebase and maybe squash, or cleanup the git history? There is some conflict.

static/usr/lib/core/rtc-sys-time-init
ernestl: Thanks, that is a good point, fixed.

static/usr/lib/core/rtc-sys-time-init
alfonsosanchezbeato, line 84: If I understand this correctly, this is setting the system time before comparing with the current system time, which might be more modern maybe if it has synced from a different clock already?

Also, maybe I am being naive, but the script seems a bit too complex, I wonder if it would be possible to instead read all possible sources of truth:
- /usr/lib/clock-epoch
- /var/lib/systemd/timesync/clock
- hw clock (can be skipped it hctosys=1?)

in a more sequential way, then take the latest, compare with current system time, and set if necessary?

static/usr/libexec/core/get-arg
```diff
@@ -0,0 +1,85 @@
...
+   cmdline="${SYSTEMD_PROC_CMDLINE}"
+else
+   cmdline=$(cat /proc/cmdline)
+fi
+
+set --
+
+# We cannot use ANSI-C quoting (e.g. $'\n') in busybox-initramfs
+# dash's printf builtin, so we make sure does not handle \x
+whitespaces="$(/usr/bin/printf '\t\n\v\f\r \xA0')"
```
alfonsosanchezbeato, line 35: Note that in the end octal can be used, see https://github.com/snapcore/core-base/pull/129

alfonsosanchezbeato (commented):
Some minor comments, looks great otherwise.

I have also one question, does the kernel change the time of the RTC when system time is set with `date -s`? If that is the case, is that done for all RTC clocks or only for the one marked with  the hctosys attribute.

static/usr/lib/core/rtc-sys-time-init
ernestl: Improved

static/usr/lib/systemd/system/rtc-sys-time-init@.service
ernestl: Improved as suggested

static/usr/libexec/core/get-arg
ernestl: Noted, thanks. 

My suggestion is that we back-port the official fix so that this diff will fall away on final merge.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,112 @@
+#!/bin/sh
+
+# Adjust system time for the following reasons:
+#     1) Sync system time to RTC time
+#          - On older kernel versions, RTC modules with HTCOSYS enabled cannot sync system time with
```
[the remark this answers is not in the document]
ernestl: Well spotted, thanks!

static/usr/lib/core/rtc-sys-time-init
ernestl: Did this and related simplification.

alfonsosanchezbeato:
I am also wondering, do we need the generator? Cannot the unit for the case when `rtc_sys_time_init` is set be generated from a udev rule as well? Just so the unit is generated from the same place always (as I understand that creating both from the generator is not possible as maybe the kernel module for the rtc has not been yet loaded).

ernestl:
> Some minor comments, looks great otherwise.
> 
> I have also one question, does the kernel change the time of the RTC when system time is set with `date -s`? If that is the case, is that done for all RTC clocks or only for the one marked with the hctosys attribute.

This script never change the time of the RTC:
   `date -s` only sets the system time
   `hwclock -r ` reads the RTC time
   `hwclock -s` sets the system time to the RTC time

systemd-timesyncd have the capability to set the RTC clock to the system time once that is trusted (e.g. after NTP sync).
This relies on CONFIG_RTC_SYSTOHC enabled and happens only for the RTC specfied RTC_HCTOSYS_DEVICE.

static/usr/lib/core/rtc-sys-time-init
ernestl: This change originates from the problem that kernel 5.4 used in some UC20 images cannot do "hctosys" sync on module loads. "hctosys" treats the RTC time as trusted and blindly syncs the system time, and perhaps this is a bit unrefined, but the intention seems to be to allow setting time backwards and forwards. 

The proposal here is that RTC clock can be used as trusted source that move system time both backward and forward. It is more refined than the RTC module logic by defining simple criteria for trusting the RTC time:  If it is more recent than the most recent timestamp. Cases like bad battery or unwanted reset would result in not using RTC time as a trusted source. It does not aim to limit adjustments to monotonic only.

In a system where system-timesycd works as expected (there are some shortcomings in version used in UC20) it can be trusted to update the clock timestamp and RTC as required to work well with this logic (and hctosys when it works). Where user/applications directly set time (as per Honeywell), they share the responsibility to do it in a sensible way.

ernestl:
> I am also wondering, do we need the generator? Cannot the unit for the case when `rtc_sys_time_init` is set be generated from a udev rule as well? Just so the unit is generated from the same place always (as I understand that creating both from the generator is not possible as maybe the kernel module for the rtc has not been yet loaded).

Thanks for the question, the issue lies with dependencies that needs to be in place early to ensure correct sequence of events: the scripts needs to run before `sysinit.target` and more specifically `system-timesyncd` which is also activated for the same target. Naturally the udev "system_wants" would invoke the service/script after `system-timesyncd`, so establishing the dependencies late does not work.

valentindavid:
> do we need the generator?

The generator's role is to introduce dependency to delay start of anything depending on `time-set.target`.

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,109 @@
...
+timesync_clock_stamp=0
+if [ -e "/var/lib/systemd/timesync/clock" ]; then
+    timesync_clock_stamp="$(stat -L "/var/lib/systemd/timesync/clock" -c '%Y')"
+else
+    journal warning "Warning: cannot stat /var/lib/systemd/timesync/clock (expected on first boot only)"
+fi
+timesync_clock_date=$(date -d @"$timesync_clock_stamp")
+journal debug "timesyncd timestamp: $timesync_clock_date"
+
+most_recent_stamp=$(echo "$clock_epoch_stamp $timesync_clock_stamp" | tr ' ' '\n' | sort -rn | head -1)
```
alfonsosanchezbeato, line 64: nitpick: we are running a few external processes for something that could be done with a comparison

static/usr/lib/core/rtc-sys-time-init
```diff
@@ -0,0 +1,109 @@
...
+    else
+        rtc_stamp=$(date -d "$rtc_stamp" +%s)
+        rtc_date=$(date -d @"$rtc_stamp")
+        echo "Successfully read RTC time: $rtc_date"
+
+        if [ "$rtc_stamp" -gt "$most_recent_stamp" ]; then
+            echo "RTC time is valid, syncing system time..."
+            if hwclock -s --noadjfile -u -f "$dev_node" 2>/dev/null; then
+                echo "Successfully synced system time"
+                exit 0
```
alfonsosanchezbeato, line 83: I find a bit strange that if the kernel has set system time from RTC we don't fully trust that and check against `$most_recent_stamp` later in the script, but if we sync because we got the kernel command line argument we do not check agains that variable too. As in the end we can have the same issue of, say, empty clock battery, in both cases.

If this is because in we know that in the second case `systemd-timesyncd` will handle that case as we know that the script will run before a comment should be added.

alfonsosanchezbeato (commented):
@ernestl thanks a lot for your detailed comments, that was really appreciated! I have a couple of minor remarks, see comments.

The other thing I wonder is about the name of the kernel argument, `rtc_sys_time_init`. We had discussions in the past as having a "namespace" for UC specific parameters to avoid them getting as params to the init process, something like `ubuntu_core.tc_sys_time_init` would be better, I think. Or having `snapd_` as prefix as we have for other kernel arguments.

static/usr/lib/core/rtc-sys-time-init
alfonsosanchezbeato: Ah, that's right, I got confused here, thanks...

ernestl:
> @ernestl thanks a lot for your detailed comments, that was really appreciated! I have a couple of minor remarks, see comments.
> 
> The other thing I wonder is about the name of the kernel argument, `rtc_sys_time_init`. We had discussions in the past as having a "namespace" for UC specific parameters to avoid them getting as params to the init process, something like `ubuntu_core.tc_sys_time_init` would be better, I think. Or having `snapd_` as prefix as we have for other kernel arguments.

For the record, this was discussed and we decided on `ubuntu_core.rtc_sys_time_init`.

static/usr/lib/core/rtc-sys-time-init
ernestl: This was improved

alfonsosanchezbeato (approved):
LGTM, thanks for the changes!

PR #151, opened 2023-07-03
get-arg: fix call to printf

valentindavid:
Anything with and "A" or a "0" in the name was not parsed correctly and interpreted as a space instead.

Backport of https://github.com/snapcore/core-base/pull/129


Meulengracht (approved):
Thank you!

PR #152, opened 2023-07-03
Remove cleanup-timesyncd hack

valentindavid:
It always removes the clock stamp so it breaks timesyncd ability to put forward the clock to a time further than when system was last powered down, when no rtc is available.

Backport of https://github.com/snapcore/core-base/pull/131


alfonsosanchezbeato (approved):
LGTM

PR #157, opened 2023-09-27
Use ucdev ppa instead of ubuntu-image

valentindavid:
secureboot-db was copied to ucdev. This will make it easier for the team to add custom packages as we need now a patched version of systemd.


mvo5 (approved):
Thank you!

alfonsosanchezbeato (approved):
LGTM!

PR #158, opened 2024-01-16
Stop running auto-import directly from udevd

valentindavid:
This is a backport of https://github.com/snapcore/core-base/pull/35 and https://github.com/snapcore/core-base/pull/165

Currently, user assertions do not work on first boot which make snapd spread test fail. The reason is because it tries to run "snap" command before snapd is mounted. The issue does not appear to be a problem on UC 22, because the service has implicit dependencies to snapd being mounted. So let's do the same on UC 20.

Original commit message:

`snap auto-import` cannot run anymore because of sandboxing of systemd-udevd. Instead we need to schedule a systemd service.

for newer core versions.

### DIFF
--- a/static/usr/lib/systemd/system/snapd.autoimport-device@.service
+++ b/static/usr/lib/systemd/system/snapd.autoimport-device@.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Auto import assertions from a specific block device
+After=snapd.service snapd.socket snapd.seeded.service
+ConditionKernelCommandLine=snapd_recovery_mode=run
+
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato:
Have the 2 mentioned PRs been backported to core22 too?

Meulengracht (approved):
This looks good to me, but also same question as alfonso

valentindavid:
I have opened https://github.com/snapcore/core-base/pull/169

alfonsosanchezbeato (approved):
LGTM, thanks

PR #162, opened 2024-04-02
static/usr/lib/tmpfiles.d: create journald config drop-in directory

Meulengracht:
Core20 version of https://github.com/snapcore/core-base/pull/200


valentindavid (approved):
Looks good. Tested, and the directory is correctly created.

When we get rid of writable path, we will probably need to prefix the path with `/writable/system-data` because of order of creation. But it is fine like that for now.

alfonsosanchezbeato (commented):
LGTM, thanks

PR #168, opened 2024-10-01
Makefile,tools: skip whitelist on LP

Meulengracht:

for newer core versions.

### DIFF
--- a/tools/generate-changelog.py
+++ b/tools/generate-changelog.py
@@ -40,7 +40,8 @@ from collections import namedtuple
 # to generate changelogs
 pkg_allowed_list = [
     'apt', # removed during hook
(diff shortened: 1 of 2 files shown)


tools/generate-changelog.py
```diff
@@ -40,7 +40,8 @@
 # to generate changelogs
 pkg_allowed_list = [
     'apt', # removed during hook
-    'debconf' # removed during hook
+    'debconf', # removed during hook
+    'ca-certificates' # no changelog in folder
```
Meulengracht, line 44:
```
snapcraft-core20 # ls parts/bootstrap/install/usr/share/doc/ca-certificates/
copyright  examples/  
```

